PR #1, opened 2020-04-01
first cut rename and integrate

trws:
Very little code change here, just header comments, file naming and directory structure.  Don't have tests tied in yet.  @crtrott, do I recall you had a makefile for that?


atomics/include/desul/atomics/Common.hpp
```diff
@@ -0,0 +1,49 @@
...
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_COMMON_HPP_
+#define DESUL_ATOMICS_COMMON_HPP_
+#include "Macros.hpp"
```
dalg24, line 11: Why do you prefer `#include "foo.hpp"` over `#include <atomics/foo.hpp>`?

atomics/include/desul/atomics/Compare_Exchange_Serial.hpp
```diff
@@ -0,0 +1,37 @@
+/* 
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+#ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_GCC_HPP_
+#define DESUL_ATOMICS_COMPARE_EXCHANGE_GCC_HPP_
```
dalg24, line 9:
```suggestion
#ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_SERIAL_HPP_
#define DESUL_ATOMICS_COMPARE_EXCHANGE_SERIAL_HPP_
```

atomics/include/desul/atomics/Common.hpp
crtrott: Generally I also would like to see "desul" somewhere in the include lines. It just feels "atomics/Compare_Exchange.hpp" is a bit generic.

trws: That was partially a mistake, I'm good with `"desul/atomics/foo.hpp"`, and will make that change shortly, but it isn't a system header or an external dependence, so I have a preference for using the more self-documenting non-system include with quotes rather than angle brackets.

trws: Actually, the reasoning for the bare "foo.hpp" was that is an unambiguous local directory include, but either way, the path version is fine.

trws:
The host side is working well it seems, but as soon as I wired in cuda I'm running into some issues.  It seems to be tied to compiling with RDC off, but anything that touches the lock array seems to break, despite having an address in the CUDA_SPACE_ATOMIC_LOCKS_DEVICE global according to the debugger.  Leaving this for a bit, but need to track that one down.

CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 ﻿cmake_minimum_required (VERSION 3.10)
-find_package(OpenMP)
+
+# Default to C++14 if not set so GTest/GMock can build
+if (NOT BLT_CXX_STD)
+    set(BLT_CXX_STD "c++14" CACHE STRING "")
+endif()
+
+include(external/blt/SetupBLT.cmake)
```
dalg24, line 8: What does BLT provide that this library need and would need be trivial to get from raw CMake?

doctest.h
```diff
@@ -0,0 +1,5971 @@
...
+// doctest.h - the lightest feature-rich C++ single-header testing framework for unit tests and TDD
+//
+// Copyright (c) 2016-2019 Viktor Kirilov
+//
+// Distributed under the MIT Software License
+// See accompanying file LICENSE.txt or copy at
+// https://opensource.org/licenses/MIT
+//
+// The documentation can be found at the library's page:
+// https://github.com/onqtam/doctest/blob/master/doc/markdown/readme.md
```
dalg24, line 14: If we are using Git submodules for BLT, why would we not to the same for doctest?

CMakeLists.txt
trws: Mainly it has reasonable handling of options for OpenMP, cuda, hip and so-forth handled for normal .cpp files without having to continually rebuild all of that from scratch. For camp I’ve admittedly avoided it so far, but it’s a quick way to bootstrap what we need for a multi-backend project and also includes gtest as a built-in option. It doesn’t necessarily have to be a sub module, heck we don’t necessarily have to use it, but as a way to make progress relatively rapidly it made sense to me to start there.

doctest.h
trws: That was actually a mistake, we don’t have anything using doctest at the moment so I’ll probably rebase that out. The direct answer to the question though is that it being a single header vendoring it directly seems reasonable, where the convention for blt is normally to do a sub module. Honestly wouldn’t mind directly vendoring that either come to think of it though. Do you have a strong preference for one way vs the other?

PR #4, opened 2020-04-14
Ensure that Kokkos was built with the requested backend

dalg24:


dalg24:
Note that in practice we will get a compile error if Kokkos was build with more than one backend enabled (besides the serial one)

Should we check for it and skip generation?

crtrott:
So if we make desul header only, why do we need to check at all? Why can't we just build against Kokkos?

trws:
I would say mainly for tests, and in case we end up with a config.h or similar at some point.  Is there a reason to keep it out? If it looks good to you @crtrott I'm fine with merging this.

crtrott:
OK I guess we can change that again.

PR #5, opened 2020-04-14
Alternative implementation for generic atomic fetch operations

dalg24:
It is undefined behavior to read from the member of the union that wasn't most recently written.

I will look into getting rid of the code duplication later if we can agree on the implementation.

### DIFF
--- a/atomics/include/desul/atomics/Generic.hpp
+++ b/atomics/include/desul/atomics/Generic.hpp
@@ -126,22 +126,27 @@ atomic_fetch_oper(const Oper& op,
                   typename std::enable_if<sizeof(T) == 4, const T>::type val,
                   MemoryOrder order,
                   MemoryScope scope) {
-  union U {
-    int32_t i;
-    T t;
-    DESUL_INLINE_FUNCTION U() {}
-  } oldval, assume, newval;
+  using fixed_width_integer_type = int32_t;
+  static_assert(sizeof(T) == sizeof(fixed_width_integer_type), "");
 
-  oldval.t = *dest;
+  T assume;
+  T newval;
+  T oldval{*dest};
 
   do {
-    assume.i = oldval.i;
-    newval.t = op.apply(assume.t, val);
-    oldval.i = desul::atomic_compare_exchange(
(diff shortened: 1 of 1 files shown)


atomics/include/desul/atomics/Generic.hpp
```diff
@@ -126,22 +126,27 @@ atomic_fetch_oper(const Oper& op,
...
+    newval = op.apply(assume, val);
+    reinterpret_cast<fixed_width_integer_type&>(oldval) =
+        desul::atomic_compare_exchange(
+            reinterpret_cast<fixed_width_integer_type* const>(dest),
+            reinterpret_cast<fixed_width_integer_type const&>(assume),
+            reinterpret_cast<fixed_width_integer_type const&>(newval),
+            order,
+            scope);
+  } while (reinterpret_cast<fixed_width_integer_type const&>(assume) !=
+           reinterpret_cast<fixed_width_integer_type const&>(oldval));
```
dalg24, line 147: Here I was tempted to just do `assume != oldval` but this adds a requirement on `T`

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -126,22 +126,27 @@ atomic_fetch_oper(const Oper& op,
...
+  T newval;
+  T oldval{*dest};
 
   do {
-    assume.i = oldval.i;
-    newval.t = op.apply(assume.t, val);
-    oldval.i = desul::atomic_compare_exchange(
-        (int32_t*)dest, assume.i, newval.i, order, scope);
-  } while (assume.i != oldval.i);
+    assume = oldval;
```
dalg24, line 137: I decided to use `T`'s assignment operator instead here.

trws (commented):
I like this direction.  For me personally I would probably reinterpret_cast each thing once into a reference to the effective type, or extract this out into a helper perhaps like we discussed on the call, rather than doing them all in place, but the functionality is the same and I agree it's worth fixing. 👍

dalg24:
Closing in favor fo #14

PR #7, opened 2020-05-21
Cuda ptx based atomics

crtrott:
Note that this only works on Volta and newer right now, due to me plugging in all the ptx assembly. I am not yet sure how to make this work also on pre-volta in a clang only build, since I can't simply exclude code via __CUDA_ARCH__>=700 …  (__device__ symbols must still exist with the same signature during the host compilation phase where __CUDA_ARCH__ is not defined, otherwise clangs internal book keeping fails).


crtrott:
Note I am trying as alternative tomorrow to just pull in the header files from CUDA 10.2 and use the functions they internally use to implement std::atomic.

atomics/include/desul/atomics/CUDA_ASM.inc
```diff
@@ -0,0 +1,920 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+================================================================================
+
+LICENSE ASSOCIATED WITH SUBSEQUENT MODIFICATIONS
+------------------------------------------------
```
dalg24, line 31: Does that mean so modifications yet?

atomics/include/desul/atomics/CUDA_CC_7.hpp
```diff
@@ -0,0 +1,161 @@
...
+DESUL_CUDA_INTEGRAL_OP_ATOMICS(MemoryOrderRelaxed, MemoryScopeDevice)
+DESUL_CUDA_INTEGRAL_OP_ATOMICS(MemoryOrderRelaxed, MemoryScopeCore)
+DESUL_CUDA_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeNode)
+DESUL_CUDA_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeDevice)
+DESUL_CUDA_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeCore)
+
+
+DESUL_CUDA_FLOAT_OP_ATOMICS(float, MemoryOrderRelaxed, MemoryScopeDevice)
+DESUL_CUDA_FLOAT_OP_ATOMICS(double, MemoryOrderRelaxed, MemoryScopeDevice)
+
```
dalg24, line 117:
```suggestion

#undef DESUL_CUDA_INTEGRAL_OP_ATOMICS
#undef DESUL_CUDA_FLOAT_OP_ATOMICS
```

atomics/include/desul/atomics/CUDA_CC_7.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include<type_traits>
+#include<desul/atomics/CUDA_ASM.hpp>
+
+#define DESUL_CUDA_INTEGRAL_OP_ATOMICS(MEMORY_ORDER, MEMORY_SCOPE)                 \
+  template <typename T>                                                           \
+  __device__									  \
+  typename std::enable_if<std::is_integral<T>::value, T>::type atomic_fetch_add(  \
+      T* const dest, T value, MEMORY_ORDER, MEMORY_SCOPE) {                       \
+    return Impl::__atomic_fetch_add_simt(dest, value, GCCMemoryOrder<MEMORY_ORDER>::value, MEMORY_SCOPE());  \
+  }                                                                               \
```
dalg24, line 20: While you are at it you could have another macro that takes `fetch_add` as argument and reuse below

atomics/include/desul/atomics/Compare_Exchange_Serial.hpp
```diff
@@ -10,6 +10,14 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #ifdef DESUL_HAVE_SERIAL_ATOMICS
 namespace desul {
+template<class MemoryScope>
+void atomic_thread_fence(MemoryOrderAcquire, MemoryScope) {
+}
+
+template<class MemoryScope>
+void atomic_thread_fence(MemoryOrderRelease, MemoryScope) {
+}
```
dalg24, line 19:
```suggestion
template<class MemoryOrder, class MemoryScope>
void atomic_thread_fence(MemoryOrder, MemoryScope) {
}
```

atomics/include/desul/atomics/CUDA_ASM.hpp
```diff
@@ -0,0 +1,115 @@
...
+        return __ATOMIC_ACQ_REL;
+      else
+        return __ATOMIC_ACQUIRE;
+    case __ATOMIC_RELEASE:
+      if (b != __ATOMIC_RELEASE)
+        return __ATOMIC_ACQ_REL;
+      else
+        return __ATOMIC_RELEASE;
+    case __ATOMIC_RELAXED: return b;
+    default: assert(0);
```
dalg24, line 67: Do we want something here when the `NDEBUG` macro is defined?

crtrott:
This only works with CC 7 and newer right now. I need to implement the fallback for CC <7. I also need to implement the PTX based compare exchange.

atomics/include/desul/atomics/cuda/cuda_cc7_asm_atomic_fetch_op.inc
```diff
@@ -0,0 +1,125 @@
+
+// Inline PTX: h u16 , r u32,  l u64, f f32, d f64
+// Ops: 
+
+// binary operations
+
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_FETCH_AND() \
+template<class ctype> \
+inline __device__ typename std::enable_if<sizeof(ctype)==4, ctype>::type atomic_fetch_and(ctype* dest, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  uint32_t asm_value = *(reinterpret_cast<uint32_t*>(&value)); \
```
dalg24, line 10:
```suggestion
  uint32_t asm_value = reinterpret_cast<uint32_t&>(value); \
```

atomics/include/desul/atomics/cuda/cuda_cc7_asm_atomic_fetch_op.inc
```diff
@@ -0,0 +1,125 @@
...
+  uint32_t asm_result = 0u; \
+  asm volatile("atom.and" __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER_ASM __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE_ASM ".b32" " %0,[%1],%2;" : "=r"(asm_result) : "l"(dest),"r"(asm_value) : "memory"); \
+  return *(reinterpret_cast<ctype*>(&asm_result)); \
+} \
+template<class ctype> \
+inline __device__ typename std::enable_if<sizeof(ctype)==8, ctype>::type atomic_fetch_and(ctype* dest, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  uint64_t asm_value = *(reinterpret_cast<uint32_t*>(&value)); \
```
dalg24, line 17:
```suggestion
  uint64_t asm_value = *(reinterpret_cast<uint64_t*>(&value)); \
```

dalg24: Error is repeated multiple times below

atomics/performance_tests/kokkos_based/PerfTestAtomicsLoc_Order.inc
```diff
@@ -0,0 +1,15 @@
+#define MEMORY_ORDER MemoryOrderRelaxed
```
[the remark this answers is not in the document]
crtrott: because I didn't know

atomics/include/desul/atomics/Common.hpp
```diff
@@ -50,4 +52,74 @@ struct MemoryScopeDevice {};
...
+  struct numeric_limits_max<uint32_t> {
+    static constexpr uint32_t value = 0xffffffffu;
+  };
+  template<>
+  struct numeric_limits_max<uint64_t> {
+    static constexpr uint64_t value = 0xfffffffflu;
+  };
+
+}
+}
```
dhollman, line 124: Can you add a comment explaining why this isn't just `std::numeric_limits`?

crtrott: Yeah. It didn't like it on CUDA without experimental const expr relaxed blah blah thingy.

dhollman: Oh right, because `max()` is a static member function and not a constant. Ugh...

atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
```diff
@@ -16,18 +16,15 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
+}
 
-template <>
-struct GCCMemoryOrder<MemoryOrderSeqCst> {
-  static constexpr int value = __ATOMIC_SEQ_CST;
-};
+template<class MemoryScope>
+void atomic_thread_fence(MemoryOrderRelease, MemoryScope) {
+  __atomic_thread_fence(__ATOMIC_RELEASE);
+}
```
dhollman, line 27: Better (?):

```
template <class MemoryOrder, class MemoryScope>
void atomic_thread_fence(MemoryOrder, MemorySpace) {
  __atomic_thread_fence(GCCMemoryOrder<MemoryOrder>::value);
}
```

atomics/include/desul/atomics/Compare_Exchange_Serial.hpp
dhollman: Oh wiat, this is for serial? I guess that's fine...

atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
crtrott: the main reason I didn't do it is so it doesn't conflict with the fully generic version, if there was one, but for atomic_thread_fence its fine I guess, since a generic version doesn't make sense. Either you hit the low level fence or you don't.

crtrott:
Regarding serial fences: these are the fallbacks for explicit non-threaded execution. This was an ask by applications which want the zero overhead option when they build with the Kokkos Serial backend for MPI only execution. Do you see any reason why we would need the memory fences to do something in that case?

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
```diff
@@ -53,6 +67,63 @@ __device__ typename std::enable_if<sizeof(T) == 8, T>::type atomic_compare_excha
                 *(reinterpret_cast<unsigned long long int*>(&value)));
   return *(reinterpret_cast<T*>(&return_val));
 }
+template <typename T, class MemoryScope>
+__device__ typename std::enable_if<sizeof(T) == 4, T>::type atomic_compare_exchange(
+    T* const dest, T compare, T value, MemoryOrderRelease, MemoryScope) {
+  static_assert(sizeof(int) == 4, "this function assumes an unsigned int is 32-bit");
```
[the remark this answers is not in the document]
crtrott: fixing all of those in my follow up PR since its comented out code anyway here.

atomics/include/desul/atomics/Common.hpp
```diff
@@ -106,6 +106,8 @@ struct CXXMemoryOrder<MemoryOrderRelease> {
 };
 }
 
+// We should in principle use std::numeric_limits, but that requires constexpr function support on device
+// Currently that is still considered experimetal on CUDA and sometimes not reliable.
```
dalg24, line 110:
```suggestion
// Currently that is still considered experimental on CUDA and sometimes not reliable.
```

atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
```diff
@@ -16,44 +16,30 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
-template <typename T, class MemoryScope>
-T atomic_compare_exchange(
-    T* dest, T compare, T value, MemoryOrderRelaxed, MemoryScope) {
-  (void)__atomic_compare_exchange_n(
-      dest, &compare, value, false, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
-  return compare;
-}
-template <typename T, class MemoryScope>
+// Failure mode for atomic_compare_exchange_n cannot be RELEASE nor ACQREL so
+// Those two get handled separatly.
```
dalg24, line 26:
```suggestion
// Those two get handled separately.
```

PR #8, opened 2020-06-04
Cuda api atomic

crtrott:
This adds the pre-volta atomics using CUDA API calls.

### DIFF
--- a/atomics/include/desul/atomics/CUDA.hpp
+++ b/atomics/include/desul/atomics/CUDA.hpp
@@ -9,6 +9,258 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_CUDA_HPP_
 
 #ifdef DESUL_HAVE_CUDA_ATOMICS
+#if defined(__CUDA_ARCH__) || !defined(__NVCC__)
+#if (__CUDA_ARCH__>=700) || !defined(__NVCC__)
 #include <desul/atomics/cuda/CUDA_asm.hpp>
 #endif
+#endif
+
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__<700)
+namespace desul {
+namespace Impl {
+template<class T>
+struct is_cuda_atomic_integer_type {
+  static constexpr bool value = std::is_same<T,int>::value ||
+                                std::is_same<T,unsigned int>::value ||
+                                std::is_same<T,unsigned long long int>::value;
+};
+
+template<class T>
+struct is_cuda_atomic_add_type {
+  static constexpr bool value = is_cuda_atomic_integer_type<T>::value ||
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 600)
+                                std::is_same<T,double>::value || 
+#endif
+                                std::is_same<T,float>::value;
+};
+
+template<class T>
+struct is_cuda_atomic_sub_type {
+  static constexpr bool value = std::is_same<T,int>::value ||
+                                std::is_same<T,unsigned int>::value;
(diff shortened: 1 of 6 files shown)


atomics/include/desul/atomics/CUDA.hpp
```diff
@@ -9,6 +9,258 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
+  T return_val = atomicAdd(dest,val);
+  __threadfence();
+  return return_val;
+}
+
+template<class T, class MemoryOrder>
+__device__ inline
+typename std::enable_if<Impl::is_cuda_atomic_add_type<T>::value,T>::type
+atomic_fetch_add(T* dest, T val, MemoryOrder, MemoryScopeCore) {
+  return atomic_fetch_add(dest,val,MemoryOrder(),MemoryScopeDevice());
```
dalg24, line 66: Why do you delegate to the `MemoryScopeDevice` overload?  Wouldn't `__threadfence_block()` be enough?

atomics/include/desul/atomics/CUDA.hpp
```diff
@@ -9,6 +9,258 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_CUDA_HPP_
 
 #ifdef DESUL_HAVE_CUDA_ATOMICS
+#if defined(__CUDA_ARCH__) || !defined(__NVCC__)
+#if (__CUDA_ARCH__>=700) || !defined(__NVCC__)
```
dalg24, line 13: Prefer
```C++
#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 700)) || !defined(__NVCC__)

File: atomics/include/desul/atomics/CUDA.hpp (hunk 1 above)
Comment (crtrott) [reply atomics/include/desul/atomics/CUDA.hpp:66]: we could do that, but need to do it conditionally on MemoryOrder. For example for relaxed, its better to call through directly, no additional memory fence is needed. Not sure how this would look for release etc.

File: atomics/include/desul/atomics/cuda/cuda_cc7_asm_atomic_op.inc [hunk 3]
```diff
@@ -8,6 +8,12 @@ inline __device__ void atomic_add(type* dest, type value, __DESUL_IMPL_CUDA_ASM_
   asm volatile("red.add" __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER_ASM __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE_ASM asm_type " [%0],%1;" :: "l"(dest),reg_type(value) : "memory"); \
 }
 
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_SUB(type,asm_type,reg_type) \
+inline __device__ void atomic_sub(type* dest, type value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  type neg_value = -value; \
```
dalg24, line 13: Maybe add a comment about unsigned int wrap around

atomics/include/desul/atomics/CUDA.hpp
trws: I think that makes sense, but it begs the question of why we don't use the atomic_thread_fence overloads to implement the fences for things like the MemoryScopeCore overloads here.  I'll put up a separate issue to take a look at it myself, there's no actual problem here since this is just more stringent than we need, but it might be nice to reclaim that bit of performance if it's cheap to implement.

PR #9, opened 2020-06-05
Fixup unit tests for non cuda build

dalg24:

PR #11, opened 2020-06-12
Add base implementation for atomic_ref

dalg24:
Will rebase when #9 gets merged in

### DIFF
--- a/atomics/include/desul/atomics/Generic.hpp
+++ b/atomics/include/desul/atomics/Generic.hpp
@@ -653,6 +653,21 @@ DESUL_INLINE_FUNCTION void atomic_dec(T* const dest,
   return atomic_sub(dest, T(1), order, scope);
 }
 
+namespace Impl {
+template <typename MemoryOrder>
+struct CmpExchFailureOrder {
+  using memory_order = std::conditional_t<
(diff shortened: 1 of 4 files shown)


atomics/include/desul/atomics/Generic.hpp
```diff
@@ -636,8 +653,40 @@ DESUL_INLINE_FUNCTION void atomic_dec(T* const dest,
...
+          class MemoryScope>
+DESUL_INLINE_FUNCTION bool atomic_compare_exchange_strong(T* const dest,
+                                                          T& expected,
+                                                          T desired,
+                                                          SuccessMemoryOrder success,
+                                                          FailureMemoryOrder failure,
+                                                          MemoryScope scope) {
+  T const old = atomic_compare_exchange(dest, expected, desired, success, scope);
+  if (old != expected) {
+    atomic_store(&expected, old, failure, scope);
```
dalg24, line 668: @crtrott I know you said it should not be an atomic store but I don't understand how `failure` gets into the play without it

crtrott: the problem is that our low level atomic_compare_exchange doesn't expose success and failure mode. But failure mode generally is weaker than success mode anyway.

PR #12, opened 2020-08-02
Fix atomics based on integration work with Kokkos

crtrott:
This has a number of fixes based on integration work with Kokkos in particular adding proper atomic_exchange and atomic_compare_exchange also for alrger types.


atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
```diff
@@ -22,29 +22,38 @@ void atomic_thread_fence(MemoryOrder, MemoryScope) {
...
   return compare;
 }
 
 template <typename T, class MemoryScope>
 T atomic_compare_exchange(
     T* dest, T compare, T value, MemoryOrderAcqRel, MemoryScope) {
-  (void)__atomic_compare_exchange_n(
-      dest, &compare, value, false, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
+  (void)__atomic_compare_exchange(
+      dest, &compare, &value, false, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
```
dalg24, line 56: Why did you decide to use `__atomic_compare_exchange`?  Not that I prefer one built-in function over the other but as far as I can tell the changes are without any effect.  Am I missing something?

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
```diff
@@ -160,5 +161,67 @@ __device__ typename std::enable_if<sizeof(T) == 8, T>::type atomic_compare_excha
...
+  // This is a way to avoid dead lock in a warp or wave front
+  T return_val;
+  int done = 0;
+  unsigned int mask = DESUL_IMPL_ACTIVEMASK;
+  unsigned int active = DESUL_IMPL_BALLOT_MASK(mask, 1);
+  unsigned int done_active = 0;
+  while (active != done_active) {
+    if (!done) {
+      if (Impl::lock_address_cuda((void*)dest, scope)) {
+        if(std::is_same<MemoryOrder,MemoryOrderSeqCst>::value) atomic_thread_fence(MemoryOrderRelease(),scope);
```
dalg24, line 180: Comment why it is ok to pay for that runtime check.

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
```diff
@@ -160,5 +161,67 @@ __device__ typename std::enable_if<sizeof(T) == 8, T>::type atomic_compare_excha
...
+DESUL_INLINE_FUNCTION 
+__device__ typename std::enable_if<(sizeof(T) != 8) && (sizeof(T) != 4), T>::type atomic_exchange(
+    T* const dest, T value, MemoryOrder, MemoryScope scope) {
+  // This is a way to avoid dead lock in a warp or wave front
+  T return_val;
+  int done = 0;
+  unsigned int mask = DESUL_IMPL_ACTIVEMASK;
+  unsigned int active = DESUL_IMPL_BALLOT_MASK(mask, 1);
+  unsigned int done_active = 0;
+  printf("Huch\n");
```
dalg24, line 205:
```suggestion
```

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -568,7 +568,8 @@ template <typename T, class MemoryOrder, class MemoryScope>
 DESUL_INLINE_FUNCTION T atomic_load(T* const dest,
                                     MemoryOrder order,
                                     MemoryScope scope) {
-  return Impl::atomic_fetch_oper(Impl::LoadOper<T, const T>(), dest, T(), order, scope);
+  using T_nc = typename std::remove_const<T>::type;
+  return Impl::atomic_fetch_oper(Impl::LoadOper<T_nc, const T_nc>(), const_cast<T_nc*>(dest), T_nc(), order, scope);
```
dalg24, line 572: I would rather take `dest` as `T const * const` and return 
`Impl::atomic_fetch_oper(Impl::LoadOper<T, const T>(), const_cast<T*>(dest), T(), order, scope);`

atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
crtrott: __atomic_compare_exchange_n doesn't work for arbitrary types. I misread that initially. Only __atomic_compare_exchange works for arbitrary types, i.e. it just works on echanging sizeof(T) using the three provided pointers.

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
crtrott: which check? That std::is_same is a compile time check in the end.

atomics/include/desul/atomics/Generic.hpp
crtrott: I don't know about that. What is the purpose of loading form a const array atomicly. Other than that you lied about it being const.

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
```diff
@@ -10,6 +10,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_LOCK_ARRAY_CUDA_HPP_
 
 #include "desul/atomics/Macros.hpp"
+#include "desul/atomics/Common.hpp"
```
dalg24, line 13: What do we need that include for?

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
dalg24: Are you saying that it gets optimized away?

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
crtrott: MemoryScope used further down. This allows me to only include the LockArrayCUDA.hpp standalone. Without all the full atomics. I need that for the Kokkos lock array integration.

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
trws: It ought to be, the value in the if is a compile time constant.  We could make all of these a call to an overloaded function instead if there's some worry about that though since it can be dispatched by type.

atomics/include/desul/atomics/Generic.hpp
trws: I think I'm with @crtrott on this one.

atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
trws: Actually, having a pre_fence/post_fence and just setting them up for dispatch might not be a bad thing for keeping the checks centralized.

PR #14, opened 2020-10-09
Make generic atomics not use undefined behavior.

crtrott:
This makes the atomic_fetch_oper and atomic_oper_fetch functions
not use undefined behavior related to unions. It also reduces
code duplication.

### DIFF
--- a/atomics/include/desul/atomics/Generic.hpp
+++ b/atomics/include/desul/atomics/Generic.hpp
@@ -136,169 +136,86 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
       ;
 }
 
-template <class Oper, typename T, class MemoryOrder, class MemoryScope>
+template<std::size_t N>
+struct atomic_compare_exchange_type;
+
+template<>
+struct atomic_compare_exchange_type<4> {
+  using type = int32_t;
+};
+
+template<>
+struct atomic_compare_exchange_type<8> {
+  using type = int64_t;
+};
+
+template<>
+struct atomic_compare_exchange_type<16> {
+  using type = Dummy16ByteValue;
+};
+
+template<class T>
+using dont_deduce_this_parameter_t = T;
+
+template <class Oper, typename T, class MemoryOrder, class MemoryScope,
+  // equivalent to:
+  //   requires atomic_always_lock_free(sizeof(T))
+  std::enable_if_t<atomic_always_lock_free(sizeof(T)), int> = 0
+>
 DESUL_INLINE_FUNCTION T
 atomic_fetch_oper(const Oper& op,
                   T* const dest,
-                  typename std::enable_if<sizeof(T) == 4, const T>::type val,
+                  dont_deduce_this_parameter_t<const T> val,
                   MemoryOrder order,
                   MemoryScope scope) {
-  union U {
-    int32_t i;
-    T t;
-    DESUL_INLINE_FUNCTION U() {}
-  } oldval, assume, newval;
-
-  oldval.t = *dest;
+  using cas_t = typename atomic_compare_exchange_type<sizeof(T)>::type;
+  cas_t oldval = reinterpret_cast<cas_t&>(*dest);
+  cas_t assume = oldval;
 
   do {
-    assume.i = oldval.i;
-    newval.t = op.apply(assume.t, val);
-    oldval.i = desul::atomic_compare_exchange(
-        (int32_t*)dest, assume.i, newval.i, order, scope);
-  } while (assume.i != oldval.i);
+    assume = oldval;
+    T newval = op.apply(reinterpret_cast<T&>(assume), val);
+    oldval = desul::atomic_compare_exchange(
+        reinterpret_cast<cas_t*>(dest), assume, reinterpret_cast<cas_t&>(newval), order, scope);
+  } while (assume != oldval);
 
-  return oldval.t;
+  return reinterpret_cast<T&>(oldval);
 }
 
-template <class Oper, typename T, class MemoryOrder, class MemoryScope>
+template <class Oper, typename T, class MemoryOrder, class MemoryScope,
+  // equivalent to:
+  //   requires atomic_always_lock_free(sizeof(T))
+  std::enable_if_t<atomic_always_lock_free(sizeof(T)), int> = 0
+>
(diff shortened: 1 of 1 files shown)


atomics/include/desul/atomics/Generic.hpp
```diff
@@ -136,169 +136,86 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
...
-  return newval.t;
-}
+template<>
+struct atomic_compare_exchange_type<16> {
+  using type = Dummy16ByteValue;
+};
 
-#if defined(DESUL_HAVE_16BYTE_COMPARE_AND_SWAP)
+template<class T>
+using dont_deduce_this_parameter_t = T;
```
dalg24, line 158:
```suggestion
template<class T>
struct dont_deduce_this_parameter { using type = T; };
template<class T>
using dont_deduce_this_parameter_t = typename dont_deduce_this_parameter<T>::type;
```

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -136,169 +136,88 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
...
-                  typename std::enable_if<sizeof(T) == 8, const T>::type val,
-                  MemoryOrder order,
-                  MemoryScope scope) {
-  union U {
-    int64_t i;
-    T t;
-    DESUL_INLINE_FUNCTION U() {}
-  } oldval, assume, newval;
+template<std::size_t N>
+struct atomic_compare_exchange_type;
```
dalg24, line 140:
```suggestion
template<std::size_t Bytes>
struct fixed_width_integer;
template<std::size_t Bytes>
using fixed_width_integer_t = typename fixed_width_integer<Bytes>::type;
```

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -382,7 +300,7 @@ atomic_oper_fetch(const Oper& op,
   }
   return return_val;
 #else
-  static_assert(false, "Unimplemented lock based attomic\n");
+  static_assert(false, "Unimplemented lock based atomic\n");
```
dalg24, line 303: You missed that one
https://github.com/desul/desul/blob/7d3b455a260f492376d0fdf9270e6cfd5c8038bf/atomics/include/desul/atomics/Generic.hpp#L255

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -136,169 +136,88 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
...
   do {
-    assume.i = oldval.i;
-    newval.t = op.apply(assume.t, val);
-    oldval.i = desul::atomic_compare_exchange(
-        (Dummy16ByteValue*)dest, assume.i, newval.i, order, scope);
-  } while (assume.i != oldval.i);
+    assume = oldval;
+    T newval = op.apply(reinterpret_cast<T&>(assume), val);
+    oldval = desul::atomic_compare_exchange(
+        reinterpret_cast<cas_t*>(dest), assume, reinterpret_cast<cas_t&>(newval), order, scope);
```
dalg24, line 181: In principle you can just do
```suggestion
        reinterpret_cast<cas_t*>(dest), assume, reinterpret_cast<cas_t>(newval), order, scope);
```
since the argument is taken by value.

I think the `reinterpret_cast<T&>(val)` was suggested as alternative to our doing `*reinterpret_cast<T*>(&val)` but here the reference is actually superfluous.

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -136,169 +136,88 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
...
                   MemoryScope scope) {
-  union U {
-    Dummy16ByteValue i;
-    T t;
-    DESUL_INLINE_FUNCTION U() {}
-  } oldval, assume, newval;
-
-  oldval.t = *dest;
+  using cas_t = typename atomic_compare_exchange_type<sizeof(T)>::type;
+  cas_t oldval = reinterpret_cast<cas_t&>(*dest);
```
dalg24, line 174:
```suggestion
  cas_t oldval = reinterpret_cast<cas_t>(*dest);
```

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -136,169 +136,88 @@ DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
...
-        (Dummy16ByteValue*)dest, assume.i, newval.i, order, scope);
-  } while (assume.i != oldval.i);
+    assume = oldval;
+    T newval = op.apply(reinterpret_cast<T&>(assume), val);
+    oldval = desul::atomic_compare_exchange(
+        reinterpret_cast<cas_t*>(dest), assume, reinterpret_cast<cas_t&>(newval), order, scope);
+  } while (assume != oldval);
 
-  return oldval.t;
+  return reinterpret_cast<T&>(oldval);
```
dalg24, line 184:
```suggestion
  return reinterpret_cast<T>(oldval);
```

atomics/include/desul/atomics/Generic.hpp
crtrott: not sure that is good idea. I mean I have my dummy 16 byte thing which clearly isn't an integer ...

atomics/include/desul/atomics/Generic.hpp
crtrott: its not since its reinterpret_cast and that can only do references or pointers if I am not mistaken ...

dalg24: Your right.  Not sure what I was thinking :)

atomics/include/desul/atomics/Generic.hpp
dalg24: `fixed_width_data` then?

trws:
I missed the existence of the "atomic_is_lock_free()" with no parameter previously.  The actual changes in here look good to me.  The name for the type seems fine, but if there's consensus to change it lets do it and get this in.  I think it's a good refactor.

dalg24 (approved):

PR #15, opened 2020-10-16
Adding OpenMP based atomics

crtrott:
First stab at adding OpenMP based atomics. Compiles and seems to work both for CPU and OpenMPTarget (for the latter as long as you don't hit the lock based atomics). Its slower though than the atomic builtins for GCC. Also note: i am using the OLD __sync_compare_and_swap GCC intrinsic for compare exchange. That does produce correct code on host and device it looks like. 

You need to explicitly add DESUL_ENABLE_OPENMP_ATOMICS to your CXX flags though

### DIFF
--- a/atomics/include/desul/atomics/Common.hpp
+++ b/atomics/include/desul/atomics/Common.hpp
@@ -141,17 +141,58 @@ using cmpexch_failure_memory_order =
 // Currently that is still considered experimetal on CUDA and sometimes not reliable.
 namespace desul {
 namespace Impl {
-  template<class T>
-  struct numeric_limits_max;
+template<class T>
+struct numeric_limits_max;
 
-  template<>
-  struct numeric_limits_max<uint32_t> {
-    static constexpr uint32_t value = 0xffffffffu;
-  };
-  template<>
-  struct numeric_limits_max<uint64_t> {
-    static constexpr uint64_t value = 0xfffffffflu;
-  };
+template<>
+struct numeric_limits_max<uint32_t> {
+  static constexpr uint32_t value = 0xffffffffu;
+};
+template<>
+struct numeric_limits_max<uint64_t> {
+  static constexpr uint64_t value = 0xfffffffflu;
+};
+
+constexpr bool atomic_always_lock_free(std::size_t size) {
+  return size == 4 || size == 8
+#if defined(DESUL_HAVE_16BYTE_COMPARE_AND_SWAP)
+         || size == 16
+#endif
+      ;
+}
+
+template <std::size_t Size, std::size_t Align>
+DESUL_INLINE_FUNCTION bool atomic_is_lock_free() noexcept {
+  return Size == 4 || Size == 8
+#if defined(DESUL_HAVE_16BYTE_COMPARE_AND_SWAP)
+         || Size == 16
+#endif
+      ;
+}
+
+template<std::size_t N>
+struct atomic_compare_exchange_type;
+
+template<>
+struct atomic_compare_exchange_type<4> {
+  using type = int32_t;
+};
+
+template<>
+struct atomic_compare_exchange_type<8> {
+  using type = int64_t;
+};
+
+template<>
+struct atomic_compare_exchange_type<16> {
+  using type = Dummy16ByteValue;
+};
+
+template<class T>
+struct dont_deduce_this_parameter { using type = T; };
+
+template<class T>
+using dont_deduce_this_parameter_t = typename dont_deduce_this_parameter<T>::type;
 
 }
 }
--- a/atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp
@@ -1,0 +1,101 @@
+/* 
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+#ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_OPENMP_HPP_
+#define DESUL_ATOMICS_COMPARE_EXCHANGE_OPENMP_HPP_
+#include "desul/atomics/Common.hpp"
+#include<cstdio>
+
+#ifdef DESUL_HAVE_OPENMP_ATOMICS
+namespace desul {
+
+#if _OPENMP > 201800
+// atomic_thread_fence for Core Scope
+inline void atomic_thread_fence(MemoryOrderSeqCst, MemoryScopeCore) {
+  #pragma omp flush acq_rel
+}
+inline void atomic_thread_fence(MemoryOrderAcqRel, MemoryScopeCore) {
+  #pragma omp flush acq_rel
+}
+inline void atomic_thread_fence(MemoryOrderRelease, MemoryScopeCore) {
+  #pragma omp flush release
+}
+inline void atomic_thread_fence(MemoryOrderAcquire, MemoryScopeCore) {
+  #pragma omp flush acquire
+}
+// atomic_thread_fence for Device Scope
+inline void atomic_thread_fence(MemoryOrderSeqCst, MemoryScopeDevice) {
+  #pragma omp flush acq_rel
+}
+inline void atomic_thread_fence(MemoryOrderAcqRel, MemoryScopeDevice) {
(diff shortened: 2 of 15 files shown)


atomics/include/desul/atomics/Common.hpp
```diff
@@ -141,17 +141,58 @@ using cmpexch_failure_memory_order =
...
+};
+
+constexpr bool atomic_always_lock_free(std::size_t size) {
+  return size == 4 || size == 8
+#if defined(DESUL_HAVE_16BYTE_COMPARE_AND_SWAP)
+         || size == 16
+#endif
+      ;
+}
+
```
crtrott, line 163: I just moved this and the following stuff.

atomics/include/desul/atomics/Common.hpp
```diff
@@ -141,17 +141,58 @@ using cmpexch_failure_memory_order =
 // Currently that is still considered experimetal on CUDA and sometimes not reliable.
 namespace desul {
 namespace Impl {
```
crtrott, line 143: white space change only

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -119,45 +119,6 @@ struct LoadOper {
   static Scalar1 apply(const Scalar1& val1, const Scalar2&) { return val1; }
 };
```
crtrott, line 121: moved this to common.hpp

atomics/include/desul/atomics/OpenMP.hpp
```diff
@@ -0,0 +1,59 @@
...
+#ifndef DESUL_ATOMICS_OPENMP_HPP_
+#define DESUL_ATOMICS_OPENMP_HPP_
+
+#ifdef DESUL_HAVE_OPENMP_ATOMICS
+
+#include<type_traits>
+
+namespace desul {
+namespace Impl {
+  template<class MEMORY_ORDER_TMP, class MEMORY_SCOPE_TMP>
```
crtrott, line 17: probably should move this to the OpenMP 40 header.

atomics/include/desul/atomics/OpenMP.hpp
```diff
@@ -0,0 +1,59 @@
...
+  }
+  template<class MEMORY_SCOPE_TMP>
+  void openmp_maybe_call_post_capture_flush(MemoryOrderSeqCst, MEMORY_SCOPE_TMP) {
+    atomic_thread_fence(MemoryOrderSeqCst(), MEMORY_SCOPE_TMP());
+  }
+
+  template<class T>
+  struct is_openmp_atomic_type_t {
+    static constexpr bool value = std::is_integral<T>::value ||
+                           std::is_floating_point<T>::value;
```
dalg24, line 50: This is the definition `std::is_arithmetic<T>::value`

crtrott:
Addressed Damiens and my own comment. And did a minimal check (relaxed add with int32_t/int64_t/uint32_t/uint64_t/float) that I get the same performance with current Kokkos atomics (which are old gcc intrinsics based) and desul atomics in OpenMPTarget.

atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include "desul/atomics/Common.hpp"
+#include<cstdio>
+
+#ifdef DESUL_HAVE_OPENMP_ATOMICS
+namespace desul {
+
+#if _OPENMP > 201800
+// atomic_thread_fence for Core Scope
+inline void atomic_thread_fence(MemoryOrderSeqCst, MemoryScopeCore) {
+  #pragma omp flush acq_rel
```
dalg24, line 19: You should comment that this is not a typo.

atomics/include/desul/atomics/openmp/OpenMP_40_op.inc
```diff
@@ -0,0 +1,101 @@
...
+  template <typename T>                                                           
+  std::enable_if_t<Impl::is_openmp_atomic_type_v<T>,T> atomic_fetch_add(  
+      T* const dest, T value, MEMORY_ORDER, MEMORY_SCOPE) {                       
+    T tmp;                                                                        
+    Impl::openmp_maybe_call_pre_capture_flush(MEMORY_ORDER(), MEMORY_SCOPE());
+    #pragma omp atomic capture                                                    
+    { tmp = *dest;  *dest += value; }                                             
+    Impl::openmp_maybe_call_post_capture_flush(MEMORY_ORDER(), MEMORY_SCOPE());   
+    return tmp;                                                                   
+  }
```
dalg24, line 11: Why did you not template on the operation?  on the memory order and scope?

crtrott: what you mean template on the operation? this is the actual overload of the function in the public API

dalg24: I spoke w/o thinking never mind.

atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp
trws: It isn't? Why isn't `seq_cst` desired in the `seq_cst` fence?

trws: I see the comment now, but `seq_cst` has been available since well before 4.0, and `acq_rel` came in after that in 5.0.  Is there a compiler that supports `acq_rel` but not `seq_cst`?

trws: And I'm wrong, OpenMP's flush doesn't have seq_cst, only atomic does.  The `seq_cst` behavior of flush is referred to as a strong flush, and is spelled `#pragma omp flush` with no memory order.  I'll follow up with some folks on why we don't let the user spell it the specific way, but that's where we are.

atomics/include/desul/atomics/Macros.hpp
```diff
@@ -43,7 +45,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_FUNCTION
 #endif
 
-#if !defined(__CUDA_ARCH__) && !defined(__HIP_DEVICE_COMPILE__)
+#if !defined(DESUL_HAVE_GPU_LIKE_PROGRESS)
```
trws, line 48: Is this a testing change? I don't see where it gets set up before this.

crtrott: line 33 or 35 its literally this definition.

crtrott: no, the macro is defined a few lines above, using those two other macros.

atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp
crtrott: ok changed that.

atomics/include/desul/atomics/Macros.hpp
trws: Got it, just skimmed past it. 🤦

atomics/include/desul/atomics/openmp/OpenMP_40.hpp
```diff
@@ -0,0 +1,97 @@
...
+// #undef MEMORY_SCOPE
+#define MEMORY_SCOPE MemoryScopeDevice
+#include<desul/atomics/openmp/OpenMP_40_op.inc>
+#undef MEMORY_SCOPE
+#define MEMORY_SCOPE MemoryScopeCore
+#include<desul/atomics/openmp/OpenMP_40_op.inc>
+#undef MEMORY_SCOPE
+#undef MEMORY_ORDER
+}  // namespace desul
+#endif
```
trws, line 97: This certainly works, but why both like this rather than letting core resolve to device through the overload?

trws (approved):
This looks good to me at this point.  I admit to wondering why the core version is split the way it is, but it doesn't change correctness so I'm good at least as a first cut.

atomics/include/desul/atomics/openmp/OpenMP_40.hpp
crtrott: we do not have a generic fallback mechanism yet. But its easy enough to change later since its isolated here.

trws: Right you are.  For some reason I remembered merging the inheritance-based fallback, but clearly that never actually happened.  This works for me, thanks!

PR #16, opened 2020-10-20
Add HIP atomics

Rombur:
This PR adds HIP atomics. There are two caveats:
1. I can't get the build system to work so I have tested using a custom `Makefile` provided by @crtrott 
2. It probably doesn't work with RDC enabled. Kokkos does not compile with RDC and thus, I can't get the tests working. Also there is no equivalent to `__CUDACC_RDC__` in HIP so I had to introduce a new macro `DESUL_HIP_RDC` but since I have trouble with the build system this macro is never defined.


atomics/include/desul/atomics/Lock_Array_HIP.hpp
```diff
@@ -0,0 +1,168 @@
...
+    ::desul::Impl::lock_array_copied = 1;                                         \
+  }
+
+#endif
+
+#if defined(DESUL_HIP_RDC) || (!defined(__HIPCC__))
+#define DESUL_ENSURE_HIP_LOCK_ARRAYS_ON_DEVICE()
+#else
+#define DESUL_ENSURE_HIP_LOCK_ARRAYS_ON_DEVICE() \
+  DESUL_IMPL_COPY_HIP_LOCK_ARRAYS_TO_DEVICE()
```
trws, line 165: Thanks for this, I'm going to do a more thorough review over the whole thing a bit later, but I think it would be good to make this a single "DESUL_ENSURE_LOCK_ARRAYS_ON_DEVICE()" macro so we don't have to keep proliferating them.  Perhaps we'll keep the individual macros, and define the common one in terms of them to keep it semi-modular, I just really don't want to end up with 4-5 of these on every offload test/region if we can help it.

crtrott (approved):
I will introduce the common ensure lock arrays macro separately.

PR #17, opened 2021-05-05
Fix incorrect usage of GCC compare exchange as well as warnings

crtrott:
We get unaligned warning for large compare exchange ops.
Also certain ops are only supported for for trivially copyable types.

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
@@ -17,13 +17,21 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #endif
 namespace desul {
 
+#if defined(__clang__) && (__clang_major__>=7) && !defined(__APPLE__)
+// Disable warning for large atomics on clang 7 and up (checked with godbolt)
+// error: large atomic operation may incur significant performance penalty [-Werror,-Watomic-alignment]
+// https://godbolt.org/z/G7YhqhbG6
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Watomic-alignment"
+#endif
 template<class MemoryOrder, class MemoryScope>
 void atomic_thread_fence(MemoryOrder, MemoryScope) {
   __atomic_thread_fence(GCCMemoryOrder<MemoryOrder>::value);
 }
 
 template <typename T, class MemoryOrder, class MemoryScope>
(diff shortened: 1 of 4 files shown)


atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
```diff
@@ -17,13 +17,20 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #endif
 namespace desul {
 
+#if defined(__clang__) && (__clang_major__>=7)
+// Disable warning for large atomics on clang 7 and up (checked with godbolt)
```
dalg24, line 21: Paste the link in the comment

atomics/include/desul/atomics/GCC.hpp
```diff
@@ -89,6 +89,43 @@ DESUL_GCC_INTEGRAL_OP_ATOMICS(MemoryOrderRelaxed, MemoryScopeCore)
 DESUL_GCC_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeNode)
 DESUL_GCC_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeDevice)
 DESUL_GCC_INTEGRAL_OP_ATOMICS(MemoryOrderSeqCst, MemoryScopeCore)
+
+template <typename T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<!std::is_trivially_copyable<T>::value, T>
```
dalg24, line 94: Why did you define these here rather than in `Compare_Exchange_GCC.hpp `

trws:
Barring @dalg24's comments, this looks good to me.

atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
crtrott: Done

atomics/include/desul/atomics/GCC.hpp
crtrott: because it actually requires the lock based stuff, and is not the fundamental compare exchange functionality which underpins all the other fallbacks.

crtrott: I guess we could forward declare the locking mechanisms ... but I think its nice that that file has only the primitives.

dalg24 (approved):

PR #18, opened 2021-05-05
Fix CUDA/HIP warnigns and make lock array functions weak symbols

crtrott:
To allow linking of external desul and internal kokkos desul I made
the lock array functions weak symbols, so that we only get one of them.

### DIFF
--- a/atomics/include/desul/atomics/HIP.hpp
+++ b/atomics/include/desul/atomics/HIP.hpp
@@ -5,8 +5,8 @@ Source: https://github.com/desul/desul
 
 SPDX-License-Identifier: (BSD-3-Clause)
 */
-#ifndef DESUL_ATOMICS_CUDA_HPP_
-#define DESUL_ATOMICS_CUDA_HPP_
+#ifndef DESUL_ATOMICS_HIP_HPP_
+#define DESUL_ATOMICS_HIP_HPP_
 
 #ifdef __HIP_DEVICE_COMPILE__
 namespace desul {
@@ -222,7 +222,7 @@ __device__ inline
 
 template <typename T, typename MemoryOrder>
 __device__ inline
-    typename std::enable_if<Impl::is_typename_atomic_integer_type<T>::value, T>::type
+    typename std::enable_if<Impl::is_hip_atomic_integer_type<T>::value, T>::type
     atomic_fetch_xor(T* dest, T val, MemoryOrder, MemoryScopeDevice) {
   __threadfence();
   T return_val = atomicXor(dest, val);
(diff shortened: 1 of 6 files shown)


atomics/include/desul/atomics/Lock_Array_Cuda.hpp
```diff
@@ -32,16 +32,23 @@ namespace Impl {
 extern int32_t* CUDA_SPACE_ATOMIC_LOCKS_DEVICE_h;
 extern int32_t* CUDA_SPACE_ATOMIC_LOCKS_NODE_h;
 
+
 /// \brief After this call, the g_host_cuda_lock_arrays variable has
 ///        valid, initialized arrays.
 ///
 /// This call is idempotent.
+/// The function is templated to make it a weak symbol to deal with Kokkos/RAJA
+///   snappshotted version while also linking against pure Desul
```
dalg24, line 41: snap[p]shoted with 2 "p"s?

atomics/include/desul/atomics/HIP.hpp
```diff
@@ -5,8 +5,8 @@ Source: https://github.com/desul/desul
 
 SPDX-License-Identifier: (BSD-3-Clause)
 */
-#ifndef DESUL_ATOMICS_CUDA_HPP_
-#define DESUL_ATOMICS_CUDA_HPP_
+#ifndef DESUL_ATOMICS_HIP_HPP_
+#define DESUL_ATOMICS_HIP_HPP_
```
dalg24, line 9: Ugggg

atomics/include/desul/atomics/HIP.hpp
```diff
@@ -261,6 +261,77 @@ __device__ inline
...
   return atomic_fetch_or(dest, val, MemoryOrder(), MemoryScopeDevice());
 }
+
+}
+
+#define DESUL_HIP_GCC_INTEGRAL_OP_ATOMICS_COMPATIBILITY(MEMORY_ORDER, MEMORY_SCOPE)                 \
+  template <typename T>                                                           \
+  __device__ typename std::enable_if<std::is_integral<T>::value && !Impl::is_hip_atomic_add_type<T>::value, T>::type atomic_fetch_add(  \
+      T* const dest, T value, MEMORY_ORDER, MEMORY_SCOPE) {                       \
+       return Impl::atomic_fetch_oper(Impl::AddOper<T, const T>(), dest, value, MEMORY_ORDER(), MEMORY_SCOPE()); \
```
dalg24, line 271: Did you consider have another macro that takes an operation and defines `atomic_fetch_oper` and `atomic_oper_fetch` and call it from your macro?  This would reduce the amount of boiler plate code further.

atomics/src/Lock_Array_HIP.cpp
```diff
@@ -38,29 +40,54 @@ namespace Impl {
 int32_t* HIP_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
 int32_t* HIP_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
 
+void check_error_and_throw_hip(hipError_t e, const std::string msg) {
```
dalg24, line 43: Define it in the anonymous namespace

atomics/src/Lock_Array_CUDA.cpp
```diff
@@ -38,18 +40,36 @@ namespace Impl {
 int32_t* CUDA_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
 int32_t* CUDA_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
 
+void check_error_and_throw_cuda(cudaError e, const std::string msg) {
```
dalg24, line 43: Define it in the anonymous namespace

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
```diff
@@ -32,16 +32,23 @@ namespace Impl {
...
+template<typename T = int>
```
dalg24, line 42:
```suggestion
template<typename /*AlwaysInt*/ = int>
```

atomics/include/desul/atomics/HIP.hpp
```diff
@@ -261,6 +261,77 @@ __device__ inline
     atomic_fetch_or(T* dest, T val, MemoryOrder, MemoryScopeCore) {
   return atomic_fetch_or(dest, val, MemoryOrder(), MemoryScopeDevice());
 }
+
+}
+
+#define DESUL_HIP_GCC_INTEGRAL_OP_ATOMICS_COMPATIBILITY(MEMORY_ORDER, MEMORY_SCOPE)                 \
```
dalg24, line 267: This addition is not captured in the description of the PR

atomics/include/desul/atomics/HIP.hpp
crtrott: yeah lol :-)

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
crtrott: because its snappy

atomics/include/desul/atomics/HIP.hpp
trws: Much as doing this with an X-macro would probably squish it down a good bit, I'm not sure I would do it right now.  Why is this here though, is it falling back from using the hip calls to the gcc builtins on hip-clang? If so, why not just always use the gcc builtins on hip, or go straight for the OpenCL/sycl support builtins so we have the full scoped set?

crtrott: so basically there is some issue with ambiguous calls made more complicated by the fact that I have __device__ and __host__ only functions for in principle different overload sets. The fully generic fallbacks are __host__ __device__, but the GCC host ones are valid for all std::is_integral, while the HIP native ones are only there for the set which is Impl::is_hip_atomic_add_type. So I needed set of __device__ only equivalents to the __host__ only equivalents provided by GCC.

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
crtrott: fixed

atomics/src/Lock_Array_HIP.cpp
crtrott: done

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
crtrott: done

atomics/src/Lock_Array_CUDA.cpp
crtrott: done

atomics/include/desul/atomics/HIP.hpp
crtrott: added to commit message

atomics/include/desul/atomics/Lock_Array_HIP.hpp
```diff
@@ -32,20 +32,22 @@ namespace Impl {
...
- * pointers, and all array memory has been freed.
- *
- * This call is idempotent.
- */
+/// \brief After this call, the g_host_cuda_lock_arrays variable has
+///        all null pointers, and all array memory has been freed.
+///
+/// This call is idempotent.
+/// The function is templated to make it a weak symbol to deal with Kokkos/RAJA
+///   snappshotted version while also linking against pure Desul
```
dalg24, line 49: Second occurrence of the same snap(p)shotted typo here

dalg24 (approved):
Any reason to delay merging this?

dalg24:
@trws please review.  We would like this to get merged in.

atomics/include/desul/atomics/HIP.hpp
trws: Ah, that makes good sense, thanks!

atomics/src/Lock_Array_CUDA.cpp
trws: Why the anonymous namespace rather than inline?

dalg24: So it has internal linkage

trws: That's kinda why I'm asking, why do we want to have multiple copies of this in a final binary rather than letting the linker throw all but one away?

dalg24: I didn't want a symbol in the library and thought the compiler was likely to decide to inline by himself.
Are you ok if add inline and keep it in the anonymous namespace?

trws (approved):

PR #19, opened 2021-05-12
Fixing some MSVC issues which cropped up during Kokkos integration

crtrott:


atomics/include/desul/atomics/Generic.hpp
```diff
@@ -43,6 +43,8 @@ struct MinOper {
   }
 };
 
+// This exit early optimization causes weird compiler errors with MSVC 2019
+#ifndef DESUL_HAVE_MSVC_ATOMICS
```
dalg24, line 47: Why didn't you just hide the specialization of `may_exit_early` then so it is always false for MSVC?

crtrott: didn't think of that. But I can try.

dalg24: Ping

atomics/include/desul/atomics/Compare_Exchange_MSVC.hpp
```diff
@@ -14,95 +15,187 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #ifndef DESUL_HAVE_16BYTE_COMPARE_AND_SWAP
 #define DESUL_HAVE_16BYTE_COMPARE_AND_SWAP
 #endif
-#include <windows.h>
 
 namespace desul {
+
+template<class T, class MemoryOrder, class MemoryScope>
+T atomic_exchange(T* const, T val, MemoryOrder, MemoryScope) { return val;}
```
dalg24, line 22: Per conversation with @trws, make sure we discuss this.

crtrott: Uh oh yeah this needs to use the same lock based fallback as below for atomic_compare_exchange

dalg24: Fixed

trws:
That is a whole lot better.  If we're doing this though, wouldn't it make sense to either use the higher_level non-intrinsic version with the appropriate memory order, [`InterlockedExchangeAcquire`](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms683594(v=vs.85)) for example, or to use the appropriate intrinsics with either the `_rel/_acq/_nf` suffixes for ARM or the `HLEAcquire/HLERelease` intrinsics for intel?  I'd be more in favor of the former and let Microsoft worry about making sure to pick the sane intrinsic for the memory order on the target given the way they do this, but as currently written even the relaxed atomics would get a full seq_cst memory fence IIUC.

crtrott:
I don't see the InterlockedExchangeAcquire in the current documentation anymore. The one you link to is listed as unmaintained for older Windows versions. Also the HLEAcquire etc are not MSVC intrinsics or? So we can certainly discuss adding a new version for Intel on Windows, but it wouldn't go into the MSVC file or?

trws:
All the intrinsics I mentioned are MSVC intrinsics, have a look [here](https://docs.microsoft.com/en-us/cpp/intrinsics/interlockedexchange-intrinsic-functions?view=msvc-160).  For some reason they offer different intrinsics on different processors, the InterlockedExchange and InterlockedExchangeAcquire/Release wrap them to do something sensible cross-platform as I understand it.

The InterlockedExchangeAcquire is apparently part of an API intended for itanium, and I see the banner saying it isn't regularly updated anymore, but it seems like that would be a much nicer way to implement this than trying to directly wrap their divergent intrinsics.  The Acquire and Release versions are still listed in their top-level current doc as the correct way to access acquire and release semantics as well, at least [here](https://docs.microsoft.com/en-us/windows/win32/sync/interlocked-variable-access).  Alternately, we could probably leverage the macros they use to map to the appropriate intrinsics in their stl [here](https://github.com/microsoft/STL/blob/472161105d596192194d4715ccad307c6c163b4a/stl/inc/xatomic.h)

trws (approved):

trws:
After sleeping on it, how likely is it people are running desul on non-x86 windows?  Made an issue to revisit, but see no reason to keep this out. Thanks @crtrott and @dalg24!

PR #20, opened 2021-06-03
added cmake dependent option for tests

JohnHynes:
The ```ENABLE_TESTS``` flag conflicts with RAJA and potentially other libraries due to current desul tests requiring kokkos.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,8 @@ endif()
 
 include(cmake/blt_boilerplate.cmake)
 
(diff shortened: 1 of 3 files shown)


trws (approved):
Makes sense to me, this is how we've been gating RAJA dependencies in general.  Any tweaks you would need to keep this in line with what you need @crtrott?

crtrott:
This is perfectly fine

trws:
Cool, thanks for the quick reply @crtrott and thanks for the PR @JohnHynes.

PR #21, opened 2021-06-10
Add SYCL atomics

masterleinad:
Matches the changes in https://github.com/kokkos/kokkos/pull/4031/files/9fee92f46596334850edcf4a099f1a76eb084c53..6597c9dc5560797c6be158fb8eb01075fd6dd104. Note that atomics for generic, i.e. large types, are not yet implemented.

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -1,0 +1,102 @@
+/*
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_SYCL_HPP_
+#define DESUL_ATOMICS_COMPARE_EXCHANGE_SYCL_HPP_
+#include "desul/atomics/Common.hpp"
+#include "desul/atomics/SYCLConversions.hpp"
+#include <CL/sycl.hpp>
+
+
+#ifdef DESUL_HAVE_SYCL_ATOMICS
+
+namespace desul {
+
+template<class MemoryOrder, class MemoryScope>
+inline void atomic_thread_fence(MemoryOrder, MemoryScope) {
+  sycl::ONEAPI::atomic_fence(DesulToSYCLMemoryOrder<MemoryOrder>::value,
+                             DesulToSYCLMemoryScope<MemoryScope>::value);
+}
+
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<sizeof(T) == 4, T>::type atomic_compare_exchange(
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned int) == 4, "this function assumes an unsigned int is 32-bit");
+  sycl::ONEAPI::atomic_ref<
+    unsigned int,
+    DesulToSYCLMemoryOrder<MemoryOrder>::value,
+    DesulToSYCLMemoryScope<MemoryScope>::value,
+    sycl::access::address_space::global_device_space>
+  dest_ref(*reinterpret_cast<unsigned int*>(dest));
+  dest_ref.compare_exchange_strong(*reinterpret_cast<unsigned int*>(&compare),
+                                   *reinterpret_cast<unsigned int*>(&value));
+  return compare;
+}
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<sizeof(T) == 8, T>::type atomic_compare_exchange(
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned long long int) == 8, "this function assumes an unsigned long long  is 64-bit");
+  sycl::ONEAPI::atomic_ref<
+    unsigned long long int,
+    DesulToSYCLMemoryOrder<MemoryOrder>::value,
+    DesulToSYCLMemoryScope<MemoryScope>::value,
+    sycl::access::address_space::global_device_space>
+  dest_ref(*reinterpret_cast<unsigned long long int*>(dest));
+  dest_ref.compare_exchange_strong(*reinterpret_cast<unsigned long long int*>(&compare),
+                                   *reinterpret_cast<unsigned long long int*>(&value));
+  return compare;
+}
+
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<sizeof(T) == 4, T>::type atomic_exchange(
+    T* const dest, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned int) == 4, "this function assumes an unsigned int is 32-bit");
+  sycl::ONEAPI::atomic_ref<
+    unsigned int,
+    DesulToSYCLMemoryOrder<MemoryOrder>::value,
+    DesulToSYCLMemoryScope<MemoryScope>::value,
+    sycl::access::address_space::global_device_space>
+  dest_ref(*reinterpret_cast<unsigned int*>(dest));
+  unsigned int return_val = dest_ref.exchange(*reinterpret_cast<unsigned int*>(&value));
+  return reinterpret_cast<T&>(return_val);
+}
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<sizeof(T) == 8, T>::type atomic_exchange(
+    T* const dest, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned long long int) == 8, "this function assumes an unsigned long long  is 64-bit");
+  sycl::ONEAPI::atomic_ref<
+    unsigned long long int,
+    DesulToSYCLMemoryOrder<MemoryOrder>::value,
+    DesulToSYCLMemoryScope<MemoryScope>::value,
+    sycl::access::address_space::global_device_space>
+  dest_ref(*reinterpret_cast<unsigned long long int*>(dest));
+  unsigned long long int return_val =
+      dest_ref.exchange(reinterpret_cast<unsigned long long int&>(value));
+  return reinterpret_cast<T&>(return_val);
+}
+
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<(sizeof(T) != 8) && (sizeof(T) != 4), T>::type atomic_compare_exchange(
+    T* const /*dest*/, T compare, T /*value*/, MemoryOrder, MemoryScope) {
+  // FIXME_SYCL not implemented
+  assert(false);
+  return compare;
+}
+
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if<(sizeof(T) != 8) && (sizeof(T) != 4), T>::type atomic_exchange(
+    T* const /*dest*/, T value, MemoryOrder, MemoryScope) {
+  // FIXME_SYCL not implemented
+  assert(false);
+  return value;
+}
+
+}
+
+#endif
+#endif
--- a/atomics/include/desul/atomics/Generic.hpp
+++ b/atomics/include/desul/atomics/Generic.hpp
@@ -247,6 +247,16 @@ atomic_fetch_oper(const Oper& op,
     done_active = DESUL_IMPL_BALLOT_MASK(done);
   }
   return return_val;
+// FIXME_SYCL not implemented
+#elif defined(__SYCL_DEVICE_ONLY__)
+  (void) op;
+  (void) dest;
+  (void) scope;
+  (void) return_val;
+  (void) done;
+
+  assert(false);
+  return val;
 #else
   unsigned int mask = DESUL_IMPL_ACTIVEMASK;
   unsigned int active = DESUL_IMPL_BALLOT_MASK(mask, 1);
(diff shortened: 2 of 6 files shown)


trws:
Glancing at this it looks good, but is the `ONEAPI` intermediary namespace necessary for this to work? Since sycl 2020 has the `sycl::atomic_ref` interface standardized I would hope that would be more portable if it's supported in DPC++.

masterleinad:
Yes, the `ONEAPI` intermediary namespace is still necessary for current releases of the `DPC++` compiler.

trws:
Darn, ok.  What would your thoughts be on having an alias or a namespace in desul that all of these refer to that we can redirect in one place then? We might well end up having to support versions with and without it, and it would be nice to centralize the choice of where that `atomic_ref` comes from.

masterleinad:
Sure. I pushed a corresponding commit.

atomics/include/desul/atomics/SYCL.hpp
```diff
@@ -0,0 +1,143 @@
...
+// Atomic Add
+template<class T, class MemoryOrder/*, class MemoryScope*/>
+inline
+typename std::enable_if<Impl::is_sycl_atomic_type<T>::value,T>::type
+atomic_fetch_add(T* dest, T val, MemoryOrder, MemoryScopeDevice) {
+  DESUL_SYCL_NAMESPACE::atomic_ref<
+    T,
+    DesulToSYCLMemoryOrder<MemoryOrder>::value,
+    DesulToSYCLMemoryScope<MemoryScopeDevice>::value,
+    sycl::access::address_space::global_device_space>
```
dalg24, line 38: Consider doing
```C++
namespace Impl {
using sycl_atomic_ref = DESUL_SYCL_NAMESPACE:: atomic_ref<T, DesulToSYCLMemoryOrder<MemoryOrder>::value, DesulToSYCLMemoryScope<MemoryScopeDevice>::value, AppropriateAddressSpace>;

File: atomics/include/desul/atomics/SYCL.hpp [hunk 2]
```diff
@@ -0,0 +1,143 @@
...
+  dest_ref(*dest);
+  return dest_ref.fetch_sub(val);
+}
+
+// Atomic Inc
+template<class MemoryOrder/*, class MemoryScope*/>
+inline
+unsigned int atomic_fetch_inc(unsigned int* dest, unsigned int val, MemoryOrder memory_order, MemoryScopeDevice memory_scope) {
+  return atomic_fetch_add(dest, val, memory_order, memory_scope);
+}
```
dalg24, line 62: Why is it not templated?  Also, I see that CUDA and HIP do that as well, but the signature makes no sense.  `val` is always `1` or we should be using `atomic_fetch_add`.

atomics/include/desul/atomics/SYCL.hpp
dalg24: Can you please comment on the address space.  I see it in the DPC++ doc but I could not find the `global_device_space` enumerator  in the SYCL specification.

dalg24 (changes requested):
Thanks for working on this

atomics/include/desul/atomics/SYCL.hpp
```diff
@@ -0,0 +1,143 @@
...
+template<class T>
+struct is_sycl_atomic_type {
+  static constexpr bool value = std::is_same<T, int>::value ||
+                                std::is_same<T, unsigned int>::value ||
+				std::is_same<T, long>::value ||
+				std::is_same<T, unsigned long>::value ||
+				std::is_same<T, long long>::value ||
+                                std::is_same<T, unsigned long long int>::value ||
+				std::is_same<T, float>::value ||
+				std::is_same<T, double>::value;
```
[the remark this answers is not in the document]
masterleinad: Sure, it would be nice if the repo would use `clang-format`

atomics/include/desul/atomics/Macros.hpp
```diff
@@ -31,7 +32,16 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
 
-#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+#ifdef SYCL_LANGUAGE_VERSION
+#define DESUL_HAVE_SYCL_ATOMICS
+#ifdef __INTEL_LLVM_COMPILER
+#define DESUL_SYCL_NAMESPACE sycl::ONEAPI
+#else
+#define DESUL_SYCL_NAMESPACE sycl
+#endif
+#endif
```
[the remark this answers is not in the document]
masterleinad: For me, it was a natural choice and I don't see much of a problem having it defined since it's prefixed with `DESUL_`.
Would you rather want to define and undefined it in all the files that use it?

atomics/include/desul/atomics/SYCL.hpp
```diff
@@ -0,0 +1,143 @@
...
+};
+} // Impl
+
+// Atomic Add
+template<class T, class MemoryOrder/*, class MemoryScope*/>
```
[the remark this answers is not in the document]
masterleinad: I would prefer to use the `SYCL` versions for all `MemoryScope`s but that doesn't work at the moment because host versions lead to ambiguous calls. I would prefer fixing/improving this behavior in a follow-up pull request such that fallback options are only used if the specialization/backend doesn't provide an implementation.

atomics/include/desul/atomics/SYCL.hpp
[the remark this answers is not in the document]
trws: We have a clang-format config in the repo.

atomics/include/desul/atomics/Generic.hpp
```diff
@@ -247,6 +247,16 @@ atomic_fetch_oper(const Oper& op,
...
+// FIXME_SYCL not implemented
+#elif defined(__SYCL_DEVICE_ONLY__)
+  (void) op;
+  (void) dest;
+  (void) scope;
+  (void) return_val;
+  (void) done;
+
+  assert(false);
+  return val;
```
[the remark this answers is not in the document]
trws: Could do that.  Since we know it isn't supported might be better to static_assert out if something tries to instantiate it if we can work out a way to do it.

atomics/include/desul/atomics/Macros.hpp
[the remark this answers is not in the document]
trws: Why a macro rather than a namespace alias or a namespace?  I was expecting something more like:

```c++
#ifdef __INTEL_LLVM_COMPILER
namespace __sycl_atomic = ::sycl::ONEAPI
#else
namespace __sycl_atomic = ::sycl
#endif
```

Or an alias for the atomic_ref if that's all that gets pulled in.

atomics/include/desul/atomics/SYCL.hpp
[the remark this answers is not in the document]
masterleinad: Which version are you using?

atomics/include/desul/atomics/Macros.hpp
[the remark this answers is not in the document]
masterleinad: Sure, I'll give it a shot.

atomics/include/desul/atomics/SYCL.hpp
masterleinad: This function has a different signature for the generic implementation that doesn't take a value (and calls `atomic_fetch_add` with `value=1`). For `CUDA`, `atomicInc` is the same as `atomicAdd` but only works on 32-bit objects.
IMHO, we should just delete this one and the corresponding ones for `CUDA` and `HIP`.

Opinions on this?

atomics/include/desul/atomics/SYCL.hpp
masterleinad: Since the current implementation only works for device scope, this seemed to be the appropriate `access_space`. The values allowed for `atomic_ref` in the release I am using right now are `global_space, local_space, global_device_space`.

The `SYCL2020` specifications say that the default value should be `generic_space` and defined values are `global_space, local_space, constant_space, private_space, generic_space`.

The values defined in the DPC++ implementation are `private_space, global_space, constant_space, local_space, global_device_space, global_host_space`.

atomics/include/desul/atomics/SYCL.hpp
trws: I actually just got someone working on fixing the generic version of that yesterday.  The inc/dec functions that take a value aren't just "add/sub one" they're "add/sub one and wrap at `val`", they have a very different use and behavior than `atomicAdd/Sub`.  The cuda and hip versions do that by default, the generic version is incomplete, and this one should either be fixed to do what the cuda and hip versions do or be removed to be fixed when the generic version is fixed.

trws: Forgot to cross-link the above to the RAJA issue where we ran into this, we expose that in RAJA for things like circular buffer management and possibly-overflowing signed counters.

llnl/raja#1073

masterleinad: Oh, you're right. I totally misread the documentation.

masterleinad: Since there are no builtins for `SYCL` for this operation, let's just fall back to the generic version and delete it here.

dalg24: Did you have to add it to get the code to compile?  If not just drop it and we can deal with CUDA and HIP elsewhere.

trws: Works for me, would you create an issue to check if one of the cl/sycl builtins in llvm might be usable for the purpose though? I can't recall and don't want to forget to at least give it a chance at better performance.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+template <typename T, class MemoryOrder, class MemoryScope>
+typename std::enable_if_t<sizeof(T) == 4, T> atomic_compare_exchange(
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned int) == 4,
+                "this function assumes an unsigned int is 32-bit");
+  sycl_atomic::atomic_ref<unsigned int,
+                          sycl_atomic::memory_order::relaxed,
+                          DesulToSYCLMemoryScope<MemoryScope>::value,
+                          sycl::access::address_space::global_device_space>
```
dalg24, line 39: Ping https://github.com/desul/desul/pull/21#discussion_r649454479

atomics/include/desul/atomics/SYCL.hpp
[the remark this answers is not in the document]
trws: What do you mean SYCL versions? These should always take the desul version, or should *at least* take the desul versions.  Do we need to come up with a scheme to avoid ambiguity due to device and host versions appearing in the same translation unit?

atomics/include/desul/atomics/SYCL.hpp
trws: This concerns me.  Are you saying that the atomics do not work on local memory or private memory? What happens if you specify generic memory?  I would want a pretty substantial warning to users if we can only support device memory atomics and not local or private atomics since we can't be sure where the pointers the functions are being used on are located.

masterleinad: The specifications say
>  For  performance  tuning  and  genericness.  Even  if  the  platform  supports  the  representation  of  thegeneric address space, this may come at some performance sacrifice. In order to help the target com­piler, it can be useful to track specifically which address space a pointer is addressing.

At the moment, the `DPC++` implementation doesn't support `generic_space`, though, (see above).

trws: That's extremely unfortunate, especially since the interface we have can't express the difference between local and global memory pointers, let alone shared memory pointers from the host.  If the only option we have is to use the global space, then so be it.  Will have to think about whether we can, or want to, do anything to be able to express this, but for now I'd say make it generic unless you're on dpc+, then whatever is least broken there.

masterleinad:
I turned out that `__INTEL_LLVM_COMPILER` isn't defined when using the open-source compiler (compiling from https://github.com/intel/llvm). Hence, I rather check for `__clang__`. (Note that I'm not aware of any compiler that would support the `sycl` namespace directly for `atomic_ref`).

atomics/include/desul/atomics/SYCL.hpp
masterleinad: See https://github.com/desul/desul/issues/29.

atomics/include/desul/atomics/SYCL.hpp
masterleinad: `dpcpp` is the only compiler I have available that supports SYCL so I am hesitant to specify something that can't be tested (despite being in the specifications).

atomics/include/desul/atomics/SYCL.hpp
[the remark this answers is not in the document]
masterleinad: I meant to say that `SYCL`'s `atomic_ref` can also be used for wider memory scopes and I would think it makes sense to use it. Defining this here for all template arguments leads to ambiguous calls, though.

atomics/include/desul/atomics/SYCL.hpp
```diff
@@ -0,0 +1,76 @@
...
+    Impl::sycl_atomic_ref<TYPE, MemoryOrder, MemoryScopeDevice> dest_ref(*dest);     \
+    dest_ref.fetch_##OPER(val);                                                      \
+    return val;                                                                      \
+  }                                                                                  \
+  template <class MemoryOrder>                                                       \
+  TYPE atomic_##OPER##_fetch(TYPE* dest, TYPE val, MemoryOrder, MemoryScopeCore) {   \
+    Impl::sycl_atomic_ref<TYPE, MemoryOrder, MemoryScopeCore> dest_ref(*dest);       \
+    dest_ref.fetch_##OPER(val);                                                      \
+    return val;                                                                      \
+  }
```
dalg24, line 42: This is not right

dalg24:
> @dalg24, any remaining issues for version one of this?

No, it is ready to merge.  This is the current version in Kokkos.

PR #23, opened 2021-06-14
Add GitHub CI

masterleinad:
This is basically taken from https://github.com/kokkos/kokkos/blob/master/.github/workflows/continuous-integration-workflow.yml to test the CPU implementation across some OS and compilers.


masterleinad:
Test results are at https://github.com/masterleinad/desul/actions/runs/936461546.

trws (approved):
This is a lot better than where we are otherwise, thanks @masterleinad!  I'll merge this, and activate checks, and we can refine from there.

PR #27, opened 2021-06-30
Fix an abbreviation in the README.md

tepperly:
DoE is the Department of Education; DOE is the Department of Energy.


trws:
Thanks @tepperly!

PR #31, opened 2021-07-26
CMake install targets

JohnHynes:
I'm working on a PR for RAJA which enables desul atomics found [here](https://github.com/LLNL/RAJA/pull/1073), and I noticed that desul has no install targets within its CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,8 @@ install(TARGETS desul_atomics
         EXPORT desul_atomics
         ARCHIVE DESTINATION lib
         LIBRARY DESTINATION lib
-        RUNTIME DESTINATION lib)
+        RUNTIME DESTINATION bin
+        INCLUDES DESTINATION include)
 
 install(DIRECTORY ${PROJECT_SOURCE_DIR}/atomics/include/
         DESTINATION include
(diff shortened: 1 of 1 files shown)


JohnHynes:
@trws Let me know if this doesn't make sense, I'm not too familiar with cmake install or which files should be included in the install for desul.

trws:
This is quite close actually.  The only things I would change are the runtime destination should pretty much always be `bin` (though we don't have anything that goes there anyway) and you can do the second part with the "INCLUDES DESTINATION" argument to install.  Have a look at this [guide](https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html) for an example, that's also probably my favorite cmake guide at the moment.  If you're willing to give it a go, adding in `DesulConfig.cmake` and `DesulTargets.cmake` files would be great too, but we can do that separately if you want to get on with other stuff.

JohnHynes:
Thank you for the guide, it seems very helpful! I mostly copied what was written in the RAJA cmake, so I'm not sure why they have ``lib`` instead of ``bin``. I gave the ``DesulConfig.cmake`` and ``DesulTargets.cmake`` a go, but I can't seem to get them to work. I'll have to move on to other things, unfortunately.

trws (approved):
Ok, thanks for this @JohnHynes!  I'll look in on RAJA's version but it's probably because raja doesn't install any executables that it slipped past us.  I'll make an issue for the config and find targets, but this is good for now.

PR #32, opened 2021-08-06
Cmake install compatibility for RAJA

JohnHynes:
Still trying to fix the compatibility of RAJA and the cmake exports for desul. This *should* fix it.

### DIFF
--- a/atomics/CMakeLists.txt
+++ b/atomics/CMakeLists.txt
@@ -23,9 +23,41 @@ endif()
 blt_add_library(NAME desul_atomics
                 HEADERS include/desul/atomics.hpp
                 ${SOURCES_ARG}
-                INCLUDES include
                 ${DEPENDS_ARG})
 
-# TODO: Add tests and install targets if needed.
+
+if(ENABLE_CUDA)
+  # Doesn't work
+  target_include_directories(desul_atomics
+    PUBLIC
+    $<INSTALL_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/atomics/cuda>)
+endif()
+
+if (ENABLE_HIP)
+  # Something
+endif()
+
+target_include_directories(desul_atomics
+  INTERFACE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>)
+
+# Install options
+install(TARGETS desul_atomics
+  EXPORT desul_atomics
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+  INCLUDES DESTINATION include)
+
+install(EXPORT desul_atomics
+  DESTINATION lib/cmake/desul)
+
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/atomics/include/
+  DESTINATION include
+  FILES_MATCHING PATTERN *.hpp)
+
+
+  # TODO: Add tests and install targets if needed.
 add_subdirectory(unit_tests)
 add_subdirectory(performance_tests)
--- a/atomics/CMakeLists.txt
+++ b/atomics/CMakeLists.txt
@@ -27,14 +27,17 @@ blt_add_library(NAME desul_atomics
 
 
 if(ENABLE_CUDA)
-  # Doesn't work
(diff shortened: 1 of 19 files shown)


JohnHynes:
@trws So the initial pass for the cmake install stuff didn't actually fix the problem with RAJA. David and I went through and made this branch as a fix.

JohnHynes:
@trws Just wanted to double-check that these changes make sense for now. They *should* let the desul integrate properly with the RAJA install stuff.

atomics/CMakeLists.txt
```diff
@@ -23,9 +23,44 @@ endif()
...
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>)
+endif()
+
+if (ENABLE_HIP)
+  target_include_directories(desul_atomics
+    PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>)
+endif()
```
trws, line 41: Why do we need these? It looks like the one below adds the same directory as these two do.

trws (commented):
Want to hear back about the duplicated BUILD/INSTALL INTERFACE items, but other than that this looks good.

atomics/CMakeLists.txt
davidbeckingsale: INTERFACE directories are only used for targets that depend on the target, but not the target itself. Without the PUBLIC include, the source files would fail to find the includes (it has to be written this way since you cannot set PUBLIC include directories on an INTERFACE target) - and desul_atomics is an interface target when built without cuda/hip.

JohnHynes:
@trws Didn't realize that there was duplicate code until after I threw up the PR, but it does allow for cuda and hip to have different install changes in the future. I think it might just make more sense to have an ``if(ENABLE_CUDA OR ENABLE_HIP)`` in there instead of duplicate if statements.

atomics/CMakeLists.txt
trws: I see now, I had missed the PUBLIC vs INTERFACE distinction.  That means it's appropriate to split these out, but should we have it as both public and interface or just one or the other?

JohnHynes:
@trws Just merged together the duplicate if-statements, does that look better?

atomics/CMakeLists.txt
```diff
@@ -23,9 +23,37 @@ endif()
 blt_add_library(NAME desul_atomics
                 HEADERS include/desul/atomics.hpp
                 ${SOURCES_ARG}
-                INCLUDES include
                 ${DEPENDS_ARG})
 
-# TODO: Add tests and install targets if needed.
+
+if(ENABLE_CUDA OR ENABLE_HIP)
```
dalg24, line 29: Why is it not unconditionally added?
What was wrong with the `INCLUDES` option with `blt_add_library`?

davidbeckingsale: Depending on whether there is a source file that will be compiled into the library or not, a different include property must be used. From my reply to Tom:

INTERFACE directories are only used for targets that depend on the target, but not the target itself. Without the PUBLIC include, the source files would fail to find the includes (it has to be written this way since you cannot set PUBLIC include directories on an INTERFACE target) - and desul_atomics is an interface target when built without cuda/hip.

davidbeckingsale: Good point! Yes, I think that `INCLUDES` for `blt_add_library` will work as long as it can take the generator expressions.

JohnHynes: Just took a peek at ``blt_add_library``, definitely seems like a better idea to use that in place of the ``target_include_directories``. Do we need to split up the source and header files into the ``SOURCES`` and ``HEADERS`` flags to achieve what we're doing right now, or does ``INCLUDES`` take care of that conflict?

JohnHynes:
@trws @dalg24 I refactored the generator expressions and they now get passed into ``blt_add_library``. Do the current changes seem appropriate? Everything seems to work with RAJA from my testing.

trws (approved):
That looks great to me, thanks @JohnHynes!

dalg24 (approved):
You might want to squash before we merge

PR #36, opened 2021-09-20
Add option to restrict GCC exchange atomics to certain sizes

dalg24:
This eliminates the need for libatomic.
Apologies for the diff, source file had not been formatted properly before.

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
@@ -17,58 +17,74 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #endif
 namespace desul {
 
-#if defined(__clang__) && (__clang_major__>=7) && !defined(__APPLE__)
+namespace Impl {
+template <class T>
+struct atomic_exchange_available_gcc {
+  constexpr static bool value =
+#ifndef DESUL_HAVE_LIBATOMIC
+      (sizeof(T) == 4 ||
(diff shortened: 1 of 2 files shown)


atomics/include/desul/atomics/Compare_Exchange_GCC.hpp
```diff
@@ -17,58 +17,74 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #endif
 namespace desul {
 
-#if defined(__clang__) && (__clang_major__>=7) && !defined(__APPLE__)
+namespace Impl {
+template <class T>
+struct atomic_exchange_available_gcc {
+  constexpr static bool value =
+#ifndef DESUL_HAVE_LIBATOMIC
```
masterleinad, line 24: Where is this defined? Is this supposed to be defined outside of `desul`?

dalg24: Outside, compiler option `-D DESUL_HAVE_LIBATOMIC=1` or in code before including desul.  (Just describing the status quo, not saying it is a great design.)

trws:
Interesting, is this to support compiling in a setup without a lib atomic runtime available?

dalg24:
> Interesting, is this to support compiling in a setup without a lib atomic runtime available?

Yes

masterleinad:
This pull request corresponds to https://github.com/kokkos/kokkos/commit/6838a66a45.

trws (approved):
Fair enough, I see no issues with this necessarily.  It's unfortunate we can't detect this from the compiler somehow though.

PR #41, opened 2021-09-21
Fix OpenMP lock-based atomic compare and exchange on host

dalg24:
Only the first commit as meaningful diff


trws (approved):
Looks good as far as I can tell.

crtrott (approved):

PR #42, opened 2021-09-23
Fixup CUDA host side function with Clang pre-volta

dalg24:


trws:
Ouch, ok.  I'm starting to seriously think we may want to get that is-a relationship between scopes and memory orders going, this is an amazing amount of duplication.  Looks like this is necessary because of clang-cuda and the requirement that every symbol on the device side is mirrored on the host side right?

crtrott:
Yes it is.

PR #45, opened 2021-10-22
Rename `atomic_wrapping_fetch_{inc,dec}` to `atomic_fetch_<TBD>_{inc,dec}`

dalg24:
Rename `atomic_wrapping_fetch_{inc,dec} -> atomic_fetch_wrapping_{inc,dec}` or something similar.

* it is more accurate because the wrapping adjectives apply to the operation "increment" or "decrement" not to "fetch".
* the current name gets in the way of efficiently using macros (see changes to HIP atomics)


trws (approved):
LGTM

masterleinad (approved):

PR #48, opened 2021-12-08
Fix SYCL atomics for local memory

masterleinad:
This pull request corresponds to https://github.com/kokkos/kokkos/pull/4585. Basically, the atomics don't do anything for `SYCL` local memory right now and this pull request introduces a check for the correct address space (if the SYCL backend is not CUDA).

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -27,13 +27,131 @@ inline void atomic_thread_fence(MemoryOrder, MemoryScope) {
       Impl::DesulToSYCLMemoryScope<MemoryScope, /*extended namespace*/ false>::value);
 }
 
+// FIXME_SYCL We need to either use generic_space or figure out how to check for the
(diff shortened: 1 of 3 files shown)


trws:
This looks fine to me, but I'm a bit leery of it without any tests. Do you have any examples we could pull in to at least get a smoke test on it?

masterleinad:
> This looks fine to me, but I'm a bit leery of it without any tests. Do you have any examples we could pull in to at least get a smoke test on it?

This doesn't do anything differently if you are not running on a non-NVIDIA GPU but we test this code at least with a NVIDIA GPU in the Kokkos CI.

trws:
Ok, then could we get a copy of one of those local memory tests to go with it?  We still need a solution to testing on intel GPUs or similar automatically, but at least it's there rather than relying on a downstream package for testing.

crtrott:
We actually don't test local_memory atomics in Kokkos proper right now. The issue showed up downstream ... we need to write testing for this I guess in general

dalg24 (approved):

crtrott (approved):

PR #49, opened 2021-12-08
Fix MSVC + NVCC macros

crtrott:


atomics/include/desul/atomics/Macros.hpp
```diff
@@ -11,14 +11,21 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
     !defined(DESUL_HAVE_SERIAL_ATOMICS)
 #define DESUL_HAVE_GCC_ATOMICS
 #endif
 
-#ifdef _MSC_VER
+// Equivalent to above: if we are compiling for the device we
+// need to use CUDA/HIP/SYCL atomics instead of MSVC atomics
+#if defined(_MSC_VER)  && \
+    (!defined(__CUDA_ARCH__) || !defined(__NVCC__)) && \
+    (!defined(__HIP_DEVICE_COMPILE) || !defined(__HIP_PLATFORM_HCC__)) && \
```
trws, line 27: Do we need the SYCL or SERIAL entries down here? If they are supposed to be essentially the same, worth making a combined macro we can check both places rather than replicating the expression?

crtrott: the SYCL is the same as the CUDA_ARCH stuff since its the device side check. We could add the SERIAL I guess to allow the SERIAL shortcut with MSVC.

atomics/include/desul/atomics/Macros.hpp
```diff
@@ -11,14 +11,21 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
+    !defined(SYCL_LANGUAGE_VERSION)
```
dalg24, line 28: The version above uses `__SYCL_DEVICE_ONLY__`.  Did you mean to use `SYCL_LANGUAGE_VERSION` here?

atomics/include/desul/atomics/Macros.hpp
crtrott: I unified the two lists

dalg24 (approved):

trws (approved):

PR #56, opened 2022-02-22
Fix bug in HIP atomics

Rombur:
There is a bug in the HIP backend. The code was adapted from the CUDA code using `NVCC` instead of the one using `CUDA-Clang`

Related to https://github.com/kokkos/kokkos/pull/4819

### DIFF
--- a/atomics/include/desul/atomics/Lock_Array_HIP.hpp
+++ b/atomics/include/desul/atomics/Lock_Array_HIP.hpp
@@ -23,6 +23,8 @@ namespace Impl {
 
 #ifdef __HIP_DEVICE_COMPILE__
 #define DESUL_IMPL_BALLOT_MASK(x) __ballot(x)
+#else
+#define DESUL_IMPL_BALLOT_MASK(x) 0
 #endif
 
 /**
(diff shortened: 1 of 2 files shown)


atomics/include/desul/atomics/Lock_Array_HIP.hpp
```diff
@@ -23,6 +23,8 @@ namespace Impl {
 
 #ifdef __HIP_DEVICE_COMPILE__
 #define DESUL_IMPL_BALLOT_MASK(x) __ballot(x)
+#else
+#define DESUL_IMPL_BALLOT_MASK(x) 0
```
dalg24, line 27: Please elaborate about that change.

dalg24 (commented):
Atomic exchange function annotation fix looks good.
What was the issue with the ballot mask macro?  It is not obvious to me.

atomics/include/desul/atomics/Lock_Array_HIP.hpp
Rombur: The macro needs to exist on the both the host and the device but you cannot call `__ballot` on the host

PR #58, opened 2022-03-23
Split atomic compare exchange for size 4/8 for HIP

crtrott:


dalg24 (changes requested):
Actually I think we should change Kokkos and keep Diesel the way it is

crtrott:
Looks like the HIP builds passed on the Kokkos side so closing this.

PR #59, opened 2022-03-23
Fix a bug in the CUDA side

crtrott:
The function is in the wrong place with respect to the ASM ifdef guard


dalg24 (approved):
This one we definitely need

trws (approved):

PR #64, opened 2022-06-02
Avoid macros for CUDA lock arrays

masterleinad:
Corresponding to https://github.com/kokkos/kokkos/pull/4682. Trying the suggestion in https://github.com/kokkos/kokkos/issues/2235.

### DIFF
--- a/atomics/include/desul/atomics/Lock_Array_Cuda.hpp
+++ b/atomics/include/desul/atomics/Lock_Array_Cuda.hpp
@@ -76,7 +76,7 @@ namespace Impl {
 /// instances in other translation units, we must update this CUDA global
 /// variable based on the Host global variable prior to running any kernels
 /// that will use it.
-/// That is the purpose of the KOKKOS_ENSURE_CUDA_LOCK_ARRAYS_ON_DEVICE macro.
+/// That is the purpose of the ensure_cuda_lock_arrays_on_device function.
 __device__
 #ifdef __CUDACC_RDC__
     __constant__ extern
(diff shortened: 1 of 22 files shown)


atomics/include/desul/atomics/Lock_Array_Cuda.hpp
```diff
@@ -138,33 +138,36 @@ namespace {
...
+  if (::desul::Impl::lock_array_copied == 0) {
+    cudaMemcpyToSymbol(::desul::Impl::CUDA_SPACE_ATOMIC_LOCKS_DEVICE,
+                       &::desul::Impl::CUDA_SPACE_ATOMIC_LOCKS_DEVICE_h,
+                       sizeof(int32_t*));
+    cudaMemcpyToSymbol(::desul::Impl::CUDA_SPACE_ATOMIC_LOCKS_NODE,
+                       &::desul::Impl::CUDA_SPACE_ATOMIC_LOCKS_NODE_h,
+                       sizeof(int32_t*));
+  }
+  ::desul::Impl::lock_array_copied = 1;
+}
```
dalg24, line 158: Why the indirection?

dalg24: By that I mean why do you bother defining `copy_cuda_lock_arrays_to_device`

dalg24 (commented):
What about HIP?

trws:
I agree with @dalg24, if this works it should probably be hoisted into the `ensure_` function, and it would be preferable for it to be the same across cuda and hip, possibly sycl if everything behaves there as well.

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
masterleinad: `copy_cuda_lock_arrays_to_device` is also used in `init_lock_arrays_cuda` and `finalize_lock_arrays_cuda`, see https://github.com/desul/desul/blob/7a85900031e382616f84d8e35ff58be5e7ded504/atomics/src/Lock_Array_CUDA.cpp#L59-L90 and is also used for the rdc case.

masterleinad:
> What about HIP?

I thought I first do `CUDA` to see if that finds agreement and then follow up with `HIP`. `SYCL` still doesn't allow having these global lock arrays (but I don't know if rdc would be a problem there anyway).

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
```diff
@@ -9,14 +9,15 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
 #include "desul/atomics/Common.hpp"
 #include "desul/atomics/Macros.hpp"
 
-#ifdef DESUL_HAVE_CUDA_ATOMICS
+namespace desul {
 
-#include <cstdint>
+#ifdef DESUL_HAVE_CUDA_ATOMICS
 
-namespace desul {
```
dalg24, line 19: This is just noise.  Please avoid doing things like that.

dalg24 (commented):
Looks fine.
Please clean-up history after you apply the changes.

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
masterleinad: `clang-format` moved the header for me and we need `ensure_cuda_lock_arrays_on_device` to be available in `namespace desul` even when `CUDA` atomics are not defined since that is used in the tests unconditionally. Of course, you could also put guards in the unit tests instead. In `Kokkos` we are using that function anyway only when compiling with CUDA support.
Anyway, I updated the pull request to minimize changes related to this now.

atomics/include/desul/atomics/Lock_Array_Cuda.hpp
dalg24: I am still confused why we have both but your changes are ok

dalg24 (approved):

trws (approved):

PR #68, opened 2022-08-24
Replace CL/sycl.hpp

masterleinad:
`CL/sycl.hpp` is deprecated in favor of `sycl/sycl.hpp`

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -13,7 +13,12 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #include "desul/atomics/SYCLConversions.hpp"
 #include "desul/atomics/Common.hpp"
 
+// FIXME_SYCL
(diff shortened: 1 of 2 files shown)


atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
```diff
@@ -13,7 +13,12 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #include "desul/atomics/SYCLConversions.hpp"
 #include "desul/atomics/Common.hpp"
 
+// FIXME_SYCL
```
dalg24, line 16: Please add a comment on the SYCL minimum version requirement

masterleinad: There is no version requirement. Older compiler only provide `CL/sycl.hpp` but the standard says this should be `sycl/sycl.hpp`. Hence, we choose the latter when the former is not available.

dalg24 (approved):

crtrott (approved):

PR #69, opened 2022-09-13
Major refactor to support compiling CUDA with NVC++

dalg24:

See kokkos/kokkos#5431

### DIFF
--- a/atomics/include/desul/atomics/Adapt_SYCL.hpp
+++ b/atomics/include/desul/atomics/Adapt_SYCL.hpp
@@ -1,0 +1,112 @@
+/*
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_ADAPT_SYCL_HPP_
+#define DESUL_ATOMICS_ADAPT_SYCL_HPP_
+
+#include <desul/atomics/Common.hpp>
+
+// FIXME_SYCL SYCL2020 dictates that <sycl/sycl.hpp> is the header to include
+// but icpx 2022.1.0 and earlier versions only provide <CL/sycl.hpp>
+#if __has_include(<sycl/sycl.hpp>)
+#include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
+#endif
+
+namespace desul {
+namespace Impl {
+
+#ifdef __clang__
+namespace sycl_sync_and_atomics = ::sycl::ext::oneapi;
+#else
+namespace sycl_sync_and_atomics = ::sycl;
+#endif
+
+template <bool extended_namespace>
+using sycl_memory_order = std::conditional_t<extended_namespace,
+                                             sycl_sync_and_atomics::memory_order,
+                                             sycl::memory_order>;
+template <bool extended_namespace>
+using sycl_memory_scope = std::conditional_t<extended_namespace,
+                                             sycl_sync_and_atomics::memory_scope,
+                                             sycl::memory_scope>;
+
+//<editor-fold desc="SYCL memory order">
+template <class MemoryOrder, bool extended_namespace = true>
+struct SYCLMemoryOrder;
+
+template <bool extended_namespace>
+struct SYCLMemoryOrder<MemoryOrderSeqCst, extended_namespace> {
+  static constexpr sycl_memory_order<extended_namespace> value =
+      sycl_memory_order<extended_namespace>::seq_cst;
+};
+template <bool extended_namespace>
+struct SYCLMemoryOrder<MemoryOrderAcquire, extended_namespace> {
+  static constexpr sycl_memory_order<extended_namespace> value =
+      sycl_memory_order<extended_namespace>::acquire;
+};
+template <bool extended_namespace>
+struct SYCLMemoryOrder<MemoryOrderRelease, extended_namespace> {
+  static constexpr sycl_memory_order<extended_namespace> value =
+      sycl_memory_order<extended_namespace>::release;
+};
+template <bool extended_namespace>
+struct SYCLMemoryOrder<MemoryOrderAcqRel, extended_namespace> {
+  static constexpr sycl_memory_order<extended_namespace> value =
+      sycl_memory_order<extended_namespace>::acq_rel;
+};
+template <bool extended_namespace>
+struct SYCLMemoryOrder<MemoryOrderRelaxed, extended_namespace> {
+  static constexpr sycl_memory_order<extended_namespace> value =
+      sycl_memory_order<extended_namespace>::relaxed;
+};
+//</editor-fold>
+
+//<editor-fold desc="SYCL memory scope">
+template <class MemoryScope, bool extended_namespace = true>
+struct SYCLMemoryScope;
+
+template <bool extended_namespace>
+struct SYCLMemoryScope<MemoryScopeCore, extended_namespace> {
+  static constexpr sycl_memory_scope<extended_namespace> value =
+      sycl_memory_scope<extended_namespace>::work_group;
+};
+
+template <bool extended_namespace>
+struct SYCLMemoryScope<MemoryScopeDevice, extended_namespace> {
+  static constexpr sycl_memory_scope<extended_namespace> value =
+      sycl_memory_scope<extended_namespace>::device;
+};
+
+template <bool extended_namespace>
+struct SYCLMemoryScope<MemoryScopeSystem, extended_namespace> {
+  static constexpr sycl_memory_scope<extended_namespace> value =
+      sycl_memory_scope<extended_namespace>::system;
+};
+//</editor-fold>
+
+// FIXME_SYCL generic_space isn't available yet for CUDA.
+#ifdef __NVPTX__
+template <class T, class MemoryOrder, class MemoryScope>
+using sycl_atomic_ref = sycl::atomic_ref<T,
+                                         SYCLMemoryOrder<MemoryOrder>::value,
+                                         SYCLMemoryScope<MemoryScope>::value,
+                                         sycl::access::address_space::global_space>;
+#else
+template <class T, class MemoryOrder, class MemoryScope>
+using sycl_atomic_ref = sycl::atomic_ref<T,
+                                         SYCLMemoryOrder<MemoryOrder>::value,
+                                         SYCLMemoryScope<MemoryScope>::value,
+                                         sycl::access::address_space::generic_space>;
+#endif
+
+}  // namespace Impl
+}  // namespace desul
+
+#endif
--- a/atomics/include/desul/atomics/Atomic_Ref.hpp
+++ b/atomics/include/desul/atomics/Atomic_Ref.hpp
@@ -9,13 +9,13 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #ifndef DESUL_ATOMIC_REF_IMPL_HPP_
 #define DESUL_ATOMIC_REF_IMPL_HPP_
 
+#include <desul/atomics/Common.hpp>
+#include <desul/atomics/Generic.hpp>
+#include <desul/atomics/Macros.hpp>
+
 #include <cstddef>
 #include <memory>
 #include <type_traits>
-
-#include "desul/atomics/Common.hpp"
-#include "desul/atomics/Generic.hpp"
-#include "desul/atomics/Macros.hpp"
 
 namespace desul {
 namespace Impl {
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -9,9 +9,9 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_SYCL_HPP_
 #define DESUL_ATOMICS_COMPARE_EXCHANGE_SYCL_HPP_
 
-// clang-format off
-#include "desul/atomics/SYCLConversions.hpp"
-#include "desul/atomics/Common.hpp"
+#include <desul/atomics/Adapt_SYCL.hpp>
+#include <desul/atomics/Common.hpp>
+#include <desul/atomics/Thread_Fence_SYCL.hpp>
 
 // FIXME_SYCL SYCL2020 dictates that <sycl/sycl.hpp> is the header to include
 // but icpx 2022.1.0 and earlier versions only provide <CL/sycl.hpp>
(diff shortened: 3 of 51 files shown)


Rombur (dismissed):
HIP good looks good

atomics/include/desul/atomics/Adapt_SYCL.hpp
```diff
@@ -0,0 +1,112 @@
...
+#endif
+
+namespace desul {
+namespace Impl {
+
+#ifdef __clang__
+namespace sycl_sync_and_atomics = ::sycl::ext::oneapi;
+#else
+namespace sycl_sync_and_atomics = ::sycl;
+#endif
```
masterleinad, line 29: We should likely try doing something less brittle. It should be fine for this pull request, but we should put a `FIXME` here linking to https://github.com/desul/desul/pull/21#issuecomment-877468175.

It also depends on the compiler versions we want to support in `desul`, of course.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
```diff
@@ -20,86 +20,76 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
   return compare;
 }
 
-template <typename T, class MemoryOrder, class MemoryScope>
-typename std::enable_if<sizeof(T) == 4, T>::type atomic_exchange(T* const dest,
-                                                                 T value,
-                                                                 MemoryOrder,
-                                                                 MemoryScope) {
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<sizeof(T) == 4, T> host_atomic_exchange(T* const dest,
```
masterleinad, line 52: Why is this called `host_atomic_exchange`? Before this pull request, we were only using the `SYCL` atomics on the device anyway.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
```diff
@@ -20,86 +20,76 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
   return compare;
 }
 
-template <typename T, class MemoryOrder, class MemoryScope>
-typename std::enable_if<(sizeof(T) != 8) && (sizeof(T) != 4), T>::type atomic_exchange(
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T> host_atomic_exchange(
     T* const /*dest*/, T value, MemoryOrder, MemoryScope) {
-  // FIXME_SYCL not implemented
-  assert(false);
```
masterleinad, line 98: Any reason to change this? It doesn't seem like trailing `FIXME` comments are consistently applied anyway.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
crtrott: but the previous overload (atomic_exchange) was available on both host and device wasn't it?

atomics/include/desul/atomics/Compare_Exchange_ScopeCaller.hpp
```diff
@@ -8,24 +8,23 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
       T* dest, T value, MEMORY_ORDER, MemoryScopeCaller) {            \
     T return_val = *dest;                                             \
     *dest = value;                                                    \
     return return_val;                                                \
   }                                                                   \
                                                                       \
-  template <typename T>                                               \
-  DESUL_INLINE_FUNCTION T atomic_compare_exchange(                    \
+  template <class T>                                                  \
+  DESUL_INLINE_FUNCTION T host_atomic_compare_exchange(               \
```
crtrott, line 27: doesn't this need a device and a host overload?

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
masterleinad: Yes, but there were some problems using it on the host.

I haven't tried to understand the strategy for the renaming of functions and it might be correct to call it `host_atomic_exchange`. I'm just asking for clarification.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
dalg24: Are you referring to moving the comment to the end of the `assert(false)` line?
(Assuming that was the question) because I find it more compact and grep-friendly.
What inconsistency are you referring to.  This is consistent at the file scope if anything.

atomics/include/desul/atomics/Compare_Exchange_ScopeCaller.hpp
dalg24: Yes you're right

atomics/include/desul/atomics/Adapt_SYCL.hpp
dalg24: > We should likely try doing something less brittle. 

I am not sure what you mean.  This is "unchanged" just being moved around.  Are you saying that we should revisit that in an upcoming PR?

> but we should put a FIXME here linking to https://github.com/desul/desul/pull/21#issuecomment-877468175.

Sure, will add.

masterleinad: > I am not sure what you mean. This is "unchanged" just being moved around. Are you saying that we should revisit that in an upcoming PR?

Yes, I'm aware that this wasn't changed and we should revisit it again (after deciding which compiler and versions to support).

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
masterleinad: Yes, I was referring to that. 
It might be that [atomics/include/desul/atomics/Compare_Exchange_OpenMP.hpp](https://github.com/desul/desul/pull/69/files#diff-e9d0a57870cabfd9e6d97723c946a7bd7f3f4c82cf1a39045725db575a32d676)
```
  // FIXME_OPENMP make sure this never gets called
  return value;
```
 is the only file where you could also move the comment to the end of the line referred to. My preference still is to have it above the respective line for readability but I can also see your point. Nothing worth holding this pull request up over.

atomics/include/desul/atomics/Thread_Fence_SYCL.hpp
```diff
@@ -0,0 +1,35 @@
...
+#endif
+
+namespace desul {
+namespace Impl {
+
+template <class MemoryOrder, class MemoryScope>
+void host_atomic_thread_fence(MemoryOrder, MemoryScope) {
+  sycl::atomic_fence(SYCLMemoryOrder<MemoryOrder, /*extended namespace*/ false>::value,
+                     SYCLMemoryScope<MemoryScope, /*extended namespace*/ false>::value);
+}
```
masterleinad, line 30: Same question here: Why is there only a `host_` version?

masterleinad:
Would you still mind summarizing the overall changes/strategy for this pull request?

atomics/include/desul/atomics/Compare_Exchange_ScopeCaller.hpp
dalg24: I went ahead and defined an `atomic_compare_exchange` overload directly (shortcutting the `host_` and `device_` version.

atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
dalg24: We were going through the SYCL atomics both on the host and the device side.  I updated so we only use them on the device side.

atomics/include/desul/atomics/Thread_Fence_SYCL.hpp
dalg24: Updated

masterleinad (dismissed):
The `SYCL` implementation looks good to me now.

atomics/include/desul/atomics/Atomic_Ref.hpp
```diff
@@ -10,13 +10,12 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMIC_REF_IMPL_HPP_
 
 #include <cstddef>
+#include <desul/atomics/Common.hpp>
+#include <desul/atomics/Generic.hpp>
+#include <desul/atomics/Macros.hpp>
```
trws, line 15: Why the change to intermixing system and project includes, is this Kokkos style?

atomics/include/desul/atomics/Fetch_Op_OpenMP.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+#ifndef DESUL_ATOMICS_FETCH_OP_OPENMP_HPP_
+#define DESUL_ATOMICS_FETCH_OP_OPENMP_HPP_
+
+#include <desul/atomics/Common.hpp>
+#include <desul/atomics/openmp/OpenMP_40.hpp>
+
+#if 0
```
trws, line 14: Is this all waiting on 5.1 support or what's the story? Can we convert the if 0 to a check on `_OPENMP` or similar perhaps?

atomics/include/desul/atomics/Fetch_Op_ScopeCaller.hpp
```diff
@@ -0,0 +1,12 @@
...
+Copyright (c) 2019, Lawrence Livermore National Security, LLC
+and DESUL project contributors. See the COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_FECH_OP_SCOPECALLER_HPP_
+#define DESUL_ATOMICS_FECH_OP_SCOPECALLER_HPP_
+
```
trws, line 11: Is this file vestigial?

atomics/include/desul/atomics/Atomic_Ref.hpp
dalg24: You mean intermixing including with double quotes and angle brackets?

atomics/include/desul/atomics/Thread_Fence_HIP.hpp
```diff
@@ -0,0 +1,35 @@
...
+#ifndef DESUL_ATOMICS_THREAD_FENCE_HIP_HPP_
+#define DESUL_ATOMICS_THREAD_FENCE_HIP_HPP_
+
+#include <desul/atomics/Common.hpp>
+
+namespace desul {
+namespace Impl {
+
+// clang-format off
+inline __device__ void device_atomic_thread_fence(MemoryOrderRelease, MemoryScopeDevice) { __threadfence();        }
```
trws, line 18: Since this and the CUDA one are character for character identical, is it worth keeping them separate?

trws (commented):
Ok, I completed a full pass.  This looks really good overall, a few questions in comments above but little to quibble about here, mainly curious about the macro guarded commented out OpenMP functions and a couple of formatting or duplication things.

atomics/include/desul/atomics/Atomic_Ref.hpp
trws: I mean mixing system libraries, parts of std for example, with parts of desul. I also notice the switch from quotes to angle brackets, but since we were using the full path rather than relative for includes that's more correct anyway unless we want to update them to use relative paths.

atomics/include/desul/atomics/Fetch_Op_ScopeCaller.hpp
dalg24: No, it was my thinking we could have shortcuts something like
```C++
template <class T>
T atomic_fetch_add(T* ptr, T val, MEMORY_ORDER, MemoryScopeCaller) {
  T ret{*ptr};
  *ptr += val
  return ret;
}
```
but I ran out of steam.  It can be removed and we will just re-introduce later if appropriate.

trws: Ok, maybe a FIXME or an issue so we can remember the intent?

atomics/include/desul/atomics/Atomic_Ref.hpp
dalg24: I tend to prefer angle brackets.  The reordering was clang-format and I also dislike it very much.  We can rework it.
I am not opposed to relative path with double-quote.  Do you want me to change?

atomics/include/desul/atomics/Fetch_Op_OpenMP.hpp
dalg24: To be honest I don't remember.  @crtrott any idea?

atomics/include/desul/atomics/Atomic_Ref.hpp
trws: Nah, fully qualified with angle brackets is fine.  I would say change the clang-format though, I tend to like these options (pseudo-syntax, I don't have a working file in front of me):

```
IncludeBlocks=Regroup
IncludeCategories:
  - Regex:           '^["<]desul/'
    Priority:        2
    CaseSensitive:   true
  - Regex:           '<[[:alnum:].]+>'
    Priority:        3
  - Regex:           '.*'
    Priority:        1
    SortPriority:    0
```

IIRC that orders unknown includes first, then desul stuff, then angle-bracket searched headers that aren't from desul last.  I'd be good with adjusting that order, but I tend to like to have project headers followed by system ones, less chance of accidental dependency.

dalg24: Any objection to fix as a follow up instead?

atomics/include/desul/atomics/Thread_Fence_HIP.hpp
dalg24: I tend to agree that it would be nice to revisit but it would raise more questions because we could have similar considerations regarding `Fetch_Op_{HIP,CUDA}.hpp` and even possibly `Compare_Exchange_{HIP,CUDA}.hpp`.
Therefore I'd rather keep them as is and revisit later as we judge fit.

atomics/include/desul/atomics/Fetch_Op_OpenMP.hpp
crtrott: I don't see where this would be coming from there is no such commented out code in the original is there?

atomics/include/desul/atomics/Fetch_Op_ScopeCaller.hpp
dalg24: Done

atomics/include/desul/atomics/Fetch_Op_OpenMP.hpp
dalg24: OK now I remember.  It was my looking into refactoring https://github.com/desul/desul/blob/497f6032effb2867e87428cbd964bcdc4183ca91/atomics/include/desul/atomics/openmp/OpenMP_40.hpp#L62-L96
and then realizing they are enough moving parts and I should probably do that as a follow up PR.
I added a FIXME.  Let me know if you prefer me to remove.

atomics/include/desul/atomics/Thread_Fence_HIP.hpp
trws: Works for me.

atomics/include/desul/atomics/Fetch_Op_OpenMP.hpp
trws: The fixme is fine, so long as we have breadcrumbs to come back.

atomics/include/desul/atomics/Atomic_Ref.hpp
trws: None, I'll create an issue from the above comment and resolve this.

trws (approved):

trws:
Excellent refactor, we might have some bits and bobs of cleanup but this is rally nice @dalg24!

PR #72, opened 2022-09-23
Fixup filename Lock_Array_{Cuda -> CUDA}.hpp

dalg24:
Messed up when cleaning up on a machine with case insensitive file system


masterleinad:
Why is this better than using the original name `Lock_Array_Cuda.hpp` and changing the include directives accordingly?

Rombur:
Because it's the only one `Cuda`, everywhere else we use `CUDA`

dalg24:
> Why is this better than using the original name `Lock_Array_Cuda.hpp` and changing the include directives accordingly?

Consistency

masterleinad:
I would favor backward compatibility, but I guess that doesn't matter for this repository yet (in particular with respect to https://github.com/desul/desul/pull/69).

trws (approved):

PR #75, opened 2022-09-27
Drop unused `{host,device}_atomic_compare_exchange_{strong,weak}`

dalg24:
The goal is to prepare for upcoming larger PR that refactors our providing host and device fallback implementation (use macros to reduce code duplication in `desul/atomics/Generic_{Device,Host}.hpp`

### DIFF
--- a/atomics/include/desul/atomics/Generic_Device.hpp
+++ b/atomics/include/desul/atomics/Generic_Device.hpp
@@ -454,42 +454,6 @@ DESUL_INLINE_FUNCTION void device_atomic_dec(T* const dest,
   return device_atomic_sub(dest, T(1), order, scope);
 }
 
-// FIXME
-template <class T,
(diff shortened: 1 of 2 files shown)


dalg24:
To be clear, I am not suggesting there is anything wrong with re-introducing later.
I am about to remove the headers `<desul/atomics/Generic_{Device,Host}.hpp>` and I don't want to sweep their removal under the carpet.
The function are just not used, as you can see from the top level API
https://github.com/desul/desul/blob/6a86952958718aee4d5969c25408a1aa07277dd2/atomics/include/desul/atomics/Generic.hpp#L371-L404
I recommend we just remove the `host` and `device` functions for now.  I expect we will revisit when cleaning up further the `<Generic.hpp>` header

trws (approved):

crtrott (approved):

PR #76, opened 2022-09-29
Refactor use macros for host and device fallback impl

dalg24:
Rational: reduced code duplication and more maintainable code
* Rename binary operators to make them more macro friendly
* Move static assertion that type is unsigned in atomic_fetch_{inc_dec}_mod API call
* Use macros to generate host and device fallback implementation


crtrott (approved):

trws (approved):
Works for me.

PR #78, opened 2022-11-10
SYCL: Implement generic atomis

masterleinad:
Corresponds to https://github.com/kokkos/kokkos/pull/5613.
The feature is described in https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_device_global.asciidoc and an extension proposed by `Intel` to the `SYCL` standard.

The feature macro is not yet implemented in the compiler, also see https://github.com/intel/llvm/pull/7340, and there are some caveats in using it (like requiring `device_image_scope`). In any case, at least the `Kokkos` unit tests work on the relevant testbeds.

The implementation here very much follows the RDC one for CUDA/HIP. For SYCL, there is no explicit compiler flag for RDC, but device functions that are called from a different translation unit must be marked with `SYCL_EXTERNAL`. Hence, there is no need to implement two different versions (and the one using `static` variables ran into some compiler/toolchain bugs).

A minor detail is that we are currently only using the `SYCL` atomics in device functions but the global variables must be visible in both compiler passes. This is somewhat hidden by using `DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED` as a guard instead of `DESUL_HAVE_SYCL_ATOMICS`.
In this pull request, it is necessary to explicitly/manually define `DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED`. There is no autodetection here but the `Kokkos` pull request has a configuration check that would set this preprocessor macro. It's up for discussion if that's something we want in this repository as well.

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -11,10 +11,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Adapt_SYCL.hpp>
 #include <desul/atomics/Common.hpp>
-// FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
-#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
 #include <desul/atomics/Lock_Array_SYCL.hpp>
-#endif
 #include <desul/atomics/Thread_Fence_SYCL.hpp>
 
 // FIXME_SYCL SYCL2020 dictates that <sycl/sycl.hpp> is the header to include
@@ -82,9 +79,7 @@ std::enable_if_t<sizeof(T) == 8, T> device_atomic_exchange(T* const dest,
 template <class T, class MemoryOrder, class MemoryScope>
 std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T>
 device_atomic_compare_exchange(
-    T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
-  // FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
-#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope scope) {
   // This is a way to avoid deadlock in a subgroup
   T return_val;
   int done = 0;
@@ -111,19 +106,11 @@ device_atomic_compare_exchange(
     done_active = group_ballot(sg, done);
   }
   return return_val;
-#else
-  (void)dest;
-  (void)value;
-  assert(false);
-  return compare;
-#endif
 }
 
 template <class T, class MemoryOrder, class MemoryScope>
 std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T> device_atomic_exchange(
-    T* const dest, T value, MemoryOrder, MemoryScope) {
-  // FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
-#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+    T* const dest, T value, MemoryOrder, MemoryScope scope) {
   // This is a way to avoid deadlock in a subgroup
   T return_val;
   int done = 0;
@@ -148,11 +135,6 @@ std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T> device_atomic_exchange
     done_active = group_ballot(sg, done);
   }
   return return_val;
-#else
-  (void)dest;
-  assert(false);
-  return value;
-#endif
 }
 
 }  // namespace Impl
--- a/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
@@ -11,6 +11,10 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Adapt_SYCL.hpp>
 #include <desul/atomics/Common.hpp>
+// FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+#include <desul/atomics/Lock_Array_SYCL.hpp>
+#endif
 #include <desul/atomics/Thread_Fence_SYCL.hpp>
 
 // FIXME_SYCL SYCL2020 dictates that <sycl/sycl.hpp> is the header to include
@@ -78,16 +82,77 @@ std::enable_if_t<sizeof(T) == 8, T> device_atomic_exchange(T* const dest,
 template <class T, class MemoryOrder, class MemoryScope>
 std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T>
 device_atomic_compare_exchange(
-    T* const /*dest*/, T compare, T /*value*/, MemoryOrder, MemoryScope) {
-  assert(false);  // FIXME_SYCL not implemented
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope scope) {
+  // FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+  // This is a way to avoid deadlock in a subgroup
+  T return_val;
+  int done = 0;
+  auto sg = sycl::ext::oneapi::experimental::this_sub_group();
+  using sycl::ext::oneapi::group_ballot;
+  using sycl::ext::oneapi::sub_group_mask;
+  sub_group_mask active = group_ballot(sg, 1);
+  sub_group_mask done_active = group_ballot(sg, 0);
+  while (active != done_active) {
+    if (!done) {
+      if (lock_address_sycl((void*)dest, scope)) {
+        if (std::is_same<MemoryOrder, MemoryOrderSeqCst>::value)
+          atomic_thread_fence(MemoryOrderRelease(), scope);
+        atomic_thread_fence(MemoryOrderAcquire(), scope);
+        return_val = *dest;
+        if (return_val == compare) {
+          *dest = value;
+          device_atomic_thread_fence(MemoryOrderRelease(), scope);
+        }
+        unlock_address_sycl((void*)dest, scope);
+        done = 1;
+      }
+    }
+    done_active = group_ballot(sg, done);
+  }
+  return return_val;
+#else
+  (void)dest;
+  (void)value;
+  assert(false);
   return compare;
+#endif
 }
 
 template <class T, class MemoryOrder, class MemoryScope>
 std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T> device_atomic_exchange(
-    T* const /*dest*/, T value, MemoryOrder, MemoryScope) {
-  assert(false);  // FIXME_SYCL not implemented
+    T* const dest, T value, MemoryOrder, MemoryScope scope) {
+  // FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+  // This is a way to avoid deadlock in a subgroup
+  T return_val;
+  int done = 0;
+  auto sg = sycl::ext::oneapi::experimental::this_sub_group();
+  using sycl::ext::oneapi::group_ballot;
+  using sycl::ext::oneapi::sub_group_mask;
+  sub_group_mask active = group_ballot(sg, 1);
+  sub_group_mask done_active = group_ballot(sg, 0);
+  while (active != done_active) {
+    if (!done) {
+      if (lock_address_sycl((void*)dest, scope)) {
+        if (std::is_same<MemoryOrder, MemoryOrderSeqCst>::value)
+          atomic_thread_fence(MemoryOrderRelease(), scope);
+        device_atomic_thread_fence(MemoryOrderAcquire(), scope);
+        return_val = *dest;
+        *dest = value;
+        device_atomic_thread_fence(MemoryOrderRelease(), scope);
+        unlock_address_sycl((void*)dest, scope);
+        done = 1;
+      }
+    }
+    done_active = group_ballot(sg, done);
+  }
+  return return_val;
+#else
+  (void)dest;
+  assert(false);
   return value;
+#endif
 }
 
 }  // namespace Impl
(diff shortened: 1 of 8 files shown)


atomics/include/desul/atomics/Compare_Exchange_SYCL.hpp
```diff
@@ -78,16 +82,77 @@ std::enable_if_t<sizeof(T) == 8, T> device_atomic_exchange(T* const dest,
 template <class T, class MemoryOrder, class MemoryScope>
 std::enable_if_t<(sizeof(T) != 8) && (sizeof(T) != 4), T>
 device_atomic_compare_exchange(
-    T* const /*dest*/, T compare, T /*value*/, MemoryOrder, MemoryScope) {
-  assert(false);  // FIXME_SYCL not implemented
+    T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
+  // FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
```
dalg24, line 87: Does this actually need to be guarded?  I feel like we could always define it and only guard in the definitions of `[un]lock_address_sycl `

atomics/src/Lock_Array_SYCL.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <sstream>
+#include <string>
+
+namespace desul {
+namespace Impl {
+SYCL_EXTERNAL
+sycl::ext::oneapi::experimental::device_global<
+    int32_t*,
+    decltype(sycl::ext::oneapi::experimental::properties(
+        sycl::ext::oneapi::experimental::device_image_scope))>
```
dalg24, line 23: What is that `device_image_scope ` property and why do we need it?

atomics/src/Lock_Array_SYCL.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace Impl {
+
+int32_t* SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
+int32_t* SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
+
+template <typename T>
+void init_lock_arrays_sycl() {
+  if (SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h != nullptr) return;
+
+  sycl::queue q;
```
dalg24, line 44: How do we know that this queue is associated with the right device?

atomics/src/Lock_Array_SYCL.cpp
dalg24: OK I saw the comment in `Lock_Array_SYCL.hpp`.  Maybe we should define some alias along these lines
```C++
template <class T>
using sycl_device_global = cl::ext::oneapi::experimental::device_global<
     T,
     decltype(sycl::ext::oneapi::experimental::properties(
         sycl::ext::oneapi::experimental::device_image_scope))>

File: atomics/src/Lock_Array_SYCL.cpp (hunk 3 above)
Comment (masterleinad) [reply atomics/src/Lock_Array_SYCL.cpp:44]: Good question! It's probably better to just pass the `sycl::queue` as an argument.

File: atomics/include/desul/atomics/Lock_Based_Fetch_Op.hpp [hunk 4]
```diff
@@ -17,7 +17,10 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #ifdef DESUL_HAVE_HIP_ATOMICS
 #include <desul/atomics/Lock_Based_Fetch_Op_HIP.hpp>
 #endif
-#ifdef DESUL_HAVE_SYCL_ATOMICS
+// FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
```
dalg24, line 21: Please confirm that this guy also need to have that special guard, not just the definition of the lock array

atomics/include/desul/atomics/Lock_Array_SYCL.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * acquiring a lock with lock_address.
+ */
+inline void unlock_address_sycl(void* ptr, desul::MemoryScopeDevice) {
+  size_t offset = size_t(ptr);
+  offset = offset >> 2;
+  offset = offset & SYCL_SPACE_ATOMIC_MASK;
+  sycl::atomic_ref<int32_t,
+                   sycl::memory_order::relaxed,
+                   sycl::memory_scope::device,
+                   sycl::access::address_space::global_space>
```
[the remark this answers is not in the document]
masterleinad: Because I know the address space in this case.

atomics/src/Lock_Array_SYCL.cpp
```diff
@@ -0,0 +1,80 @@
...
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+// FIXME_SYCL Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL when available instead
+#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+
+#include <cinttypes>
+#include <desul/atomics/Lock_Array_SYCL.hpp>
+#include <sstream>
+#include <string>
```
dalg24, line 15:
```suggestion
```

dalg24 (changes requested):
```diff
diff --git a/atomics/CMakeLists.txt b/atomics/CMakeLists.txt
index 901c992..e766e04 100644
--- a/atomics/CMakeLists.txt
+++ b/atomics/CMakeLists.txt
@@ -25,6 +25,8 @@ endif()
 option(ENABLE_SYCL "Enable SYCL support" OFF) # option does not actually yet exist on the BLT side
 if(ENABLE_SYCL)
   set(DESUL_ATOMICS_ENABLE_SYCL ON)
+  set(DESUL_SYCL_ATOMICS_SOURCES src/Lock_Array_SYCL.cpp)
+  set(SOURCES_ARG ${DESUL_SYCL_ATOMICS_SOURCES})
 endif()
 if(ENABLE_OPENMP)
   set(DESUL_ATOMICS_ENABLE_OPENMP ON)

File: atomics/include/desul/atomics/Lock_Array_SYCL.hpp [hunk 7]
```diff
@@ -0,0 +1,147 @@
...
+  offset = offset >> 2;
+  offset = offset & SYCL_SPACE_ATOMIC_MASK;
+  sycl::atomic_ref<int32_t,
+                   sycl::memory_order::relaxed,
+                   sycl::memory_scope::system,
+                   sycl::access::address_space::global_space>
+      lock_node_ref(SYCL_SPACE_ATOMIC_LOCKS_NODE[offset]);
+  lock_node_ref.exchange(0);
+}
+#else
```
dalg24, line 130: What about `{init,finalize}_locks_array_sycl`?
Is the `#ifdef DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED` misplaced?

atomics/include/desul/atomics/Lock_Array_SYCL.hpp
```diff
@@ -0,0 +1,147 @@
...
+inline bool lock_address_sycl(void*, MemoryScopeDevice) {
+  assert(false);
+  return true;
```
dalg24, line 133: Returning true so the CAS loops don't hang?
Might be worth a comment.

dalg24:
Also
```diff
diff --git a/atomics/include/desul/atomics/Lock_Array.hpp b/atomics/include/desul/atomics/Lock_Array.hpp
index a5af4c4..51e398a 100644
--- a/atomics/include/desul/atomics/Lock_Array.hpp
+++ b/atomics/include/desul/atomics/Lock_Array.hpp
@@ -49,6 +49,10 @@ inline void init_lock_arrays() {
 #ifdef DESUL_HAVE_HIP_ATOMICS
   init_lock_arrays_hip();
 #endif
+
+#ifdef DESUL_HAVE_SYCL_ATOMICS
+  init_lock_arrays_sycl();
+#endif
 }

 inline void finalize_lock_arrays() {
@@ -59,6 +63,10 @@ inline void finalize_lock_arrays() {
 #ifdef DESUL_HAVE_HIP_ATOMICS
   finalize_lock_arrays_hip();
 #endif
+
+#ifdef DESUL_HAVE_SYCL_ATOMICS
+  finalize_lock_arrays_sycl();
+#endif
 }

 inline void ensure_lock_arrays_on_device() {

**Comment by masterleinad** [comment]
`init_lock_arrays` can't really call `init_lock_arrays_sycl(sycl::queue);` since the latter takes a `sycl::queue`. I'm not sure if it's worth defining a static `queue` somewhere for that.

**Comment by dalg24** [comment]
> `init_lock_arrays` can't really call `init_lock_arrays_sycl(sycl::queue);` since the latter takes a `sycl::queue`. I'm not sure if it's worth defining a static `queue` somewhere for that.

Oh that's right

File: atomics/include/desul/atomics/Lock_Array_SYCL.hpp (hunk 7 above)
Comment (masterleinad) [reply atomics/include/desul/atomics/Lock_Array_SYCL.hpp:130]: `init_lock_arrays_sycl` and `finalize_lock_arrays_sycl` need to be called from the outside and we give `desul` the information if `device` global variables are supported (whereas `[un]lock_address_sycl` is only called internally). Thus, the call site already knows that `[init|finalize]_lock_arrays_sycl` can't be called.
Let me add a fallback implementation for good measure anyway.

File: atomics/src/Lock_Array_SYCL.cpp [hunk 9]
```diff
@@ -0,0 +1,78 @@
...
+  sycl::free(SYCL_SPACE_ATOMIC_LOCKS_NODE_h, q);
+  SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
+  SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
+}
+
+template void init_lock_arrays_sycl<int>(sycl::queue);
+template void finalize_lock_arrays_sycl<int>(sycl::queue);
+
+}  // namespace Impl
+}  // namespace desul
```
dalg24, line 77: Do you mind trying
```suggestion
SYCL_EXTERNAL
sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_DEVICE;
SYCL_EXTERNAL
sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_NODE;

int32_t* desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
int32_t* desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;

template <>
void desul::Impl::init_lock_arrays_sycl<int>(sycl::queue q) {
  if (SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h != nullptr) return;

  SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h =
      sycl::malloc_device<int32_t>(SYCL_SPACE_ATOMIC_MASK + 1, q);
  SYCL_SPACE_ATOMIC_LOCKS_NODE_h =
      sycl::malloc_host<int32_t>(SYCL_SPACE_ATOMIC_MASK + 1, q);

  // FIXME_SYCL Once supported, the following should be replaced by
  // q.memcpy(SYCL_SPACE_ATOMIC_LOCKS_DEVICE,
  //          &SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h,
  //          sizeof(int32_t*));
  // q.memcpy(SYCL_SPACE_ATOMIC_LOCKS_NODE,
  //          &SYCL_SPACE_ATOMIC_LOCKS_NODE_h,
  //          sizeof(int32_t*));
  auto device_ptr = SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h;
  auto node_ptr = SYCL_SPACE_ATOMIC_LOCKS_NODE_h;
  q.single_task([=] {
    SYCL_SPACE_ATOMIC_LOCKS_DEVICE.get() = device_ptr;
    SYCL_SPACE_ATOMIC_LOCKS_NODE.get() = node_ptr;
  });

  q.memset(SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h,
           0,
           sizeof(int32_t) * (SYCL_SPACE_ATOMIC_MASK + 1));
  q.memset(SYCL_SPACE_ATOMIC_LOCKS_NODE_h,
           0,
           sizeof(int32_t) * (SYCL_SPACE_ATOMIC_MASK + 1));

  q.wait_and_throw();
}

template <>
void desul::Impl::finalize_lock_arrays_sycl<int>(sycl::queue q) {
  if (SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h == nullptr) return;

  sycl::free(SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h, q);
  sycl::free(SYCL_SPACE_ATOMIC_LOCKS_NODE_h, q);
  SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
  SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
}
```

dalg24 (commented):
Did you mean to change blt?

atomics/src/Lock_Array_SYCL.cpp
masterleinad: It also needs
```
diff --git a/atomics/src/Lock_Array_SYCL.cpp b/atomics/src/Lock_Array_SYCL.cpp
index c925b5e..75e6205 100644
--- a/atomics/src/Lock_Array_SYCL.cpp
+++ b/atomics/src/Lock_Array_SYCL.cpp
@@ -13,9 +13,9 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #include <desul/atomics/Lock_Array_SYCL.hpp>
 
 SYCL_EXTERNAL
-sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_DEVICE;
+desul::Impl::sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_DEVICE;
 SYCL_EXTERNAL
-sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_NODE;
+desul::Impl::sycl_device_global<int32_t*> desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_NODE;
 
 int32_t* desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
 int32_t* desul::Impl::SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
```
I would still prefer using `namespace desul::Impl` than fully-qualifying everything.

trws:
> `init_lock_arrays` can't really call `init_lock_arrays_sycl(sycl::queue);` since the latter takes a `sycl::queue`. I'm not sure if it's worth defining a static `queue` somewhere for that.

My understanding is that there is a default available from OneAPI, though not SYCL portable. Is that correct?  If so, it might be worth supporting that as the value while also having the version that takes a queue for more portable use.

trws:
@Homerdin I know you're still indisposed, but you may wish to be aware of this. Should clear some of the remaining atomics for the sycl port.

dalg24 (approved):

trws (approved):
Good stuff, thanks @masterleinad!

PR #87, opened 2023-01-13
Make building tests optional

masterleinad:
Currently, it's not possible to configure `desul` without configuring tests (which requires `Kokkos`) although there is a `CMake` option that should control it.

### DIFF
--- a/atomics/CMakeLists.txt
+++ b/atomics/CMakeLists.txt
@@ -87,5 +87,7 @@ install(DIRECTORY
   PATTERN *.inc
   PATTERN *.inc_*)
 
-add_subdirectory(unit_tests)
-add_subdirectory(performance_tests)
+if(BUILD_TESTING)
+  add_subdirectory(unit_tests)
+  add_subdirectory(performance_tests)
+endif()


atomics/CMakeLists.txt
```diff
@@ -87,5 +87,7 @@ install(DIRECTORY
   PATTERN *.inc
   PATTERN *.inc_*)
 
-add_subdirectory(unit_tests)
-add_subdirectory(performance_tests)
+if(BUILD_TESTING)
```
dalg24, line 90: Please discuss the choice for the option to enable testing

dalg24: The CTest module defines a `BUILD_TESTING` but it is not obvious to me where the `include(CTest)` would occur.

dalg24: Considering https://github.com/LLNL/blt/blob/8c229991e65e7a9603c621b47cb3ba158bb7468c/cmake/BLTOptions.cmake#L16 and the other options we currently accept, is `ENABLE_TESTS` a better choice?

dalg24: And there is `DESUL_ENABLE_TESTS` as well  
https://github.com/desul/desul/blob/f63c5ee63a2f3aababaad995561d15b2b4951611/CMakeLists.txt#L14
https://github.com/desul/desul/blob/f63c5ee63a2f3aababaad995561d15b2b4951611/atomics/unit_tests/kokkos_based/CMakeLists.txt#L1

dalg24 (changes requested):
> Currently, it's not possible to configure desul without configuring tests (which requires Kokkos) although there is a CMake option that should control it.

Not true.  You can configure with `-DENABLE_TESTS=OFF` or even `-DDESUL_ENABLE_TESTS=OFF`

masterleinad:
> Not true. You can configure with `-DENABLE_TESTS=OFF` or even `-DDESUL_ENABLE_TESTS=OFF`

Ah, my bad, you're right. I was looking in the wrong place to find the right options. It might still be sensible to have testing off by default to not be forced to find a Kokkos installation by default.

PR #89, opened 2023-01-20
Drop (unused) `Impl::eliminate_warning_for_lock_array` CUDA/HIP functions

dalg24:
Corresponding to kokkos/kokkos#5786

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -33,14 +33,14 @@ jobs:
         run: |
           git submodule init
           git submodule update
-     - name: Checkout Kokkos
+      - name: Checkout Kokkos
         uses: actions/checkout@v2.2.0
         with:
           repository: kokkos/kokkos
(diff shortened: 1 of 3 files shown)


masterleinad (commented):
Looks good but have you actually tested that there is no warning?

dalg24:
> Looks good but have you actually tested that there is no warning?

What warning?

dalg24:
Rebased onto master after the merge of #90

masterleinad:
> What warning?

It's called `eliminate_warning_for_lock_array` so I suspect there was a warning at some point. 🙂 SInce we are not running unit tests with `Cuda` or `HIP` in the CI, I just want to make sure.

dalg24:
These were clearly copied from Kokkos
https://github.com/kokkos/kokkos/blob/6e73a35407bba14313d30f420f716cb91b566e7d/core/src/Cuda/Kokkos_Cuda_Locks.hpp#L122
were they are also unused :/

dalg24:
Didn't mean to close...

PR #88, opened 2023-01-17
Replace HIP_LOCK_ARRAYS macros by functions

masterleinad:
Corresponding to https://github.com/kokkos/kokkos/pull/5770. Analogous to https://github.com/desul/desul/pull/64.

PR #90, opened 2023-01-20
Test with Kokkos >=4.0.0 commit instead of 3.4.01

masterleinad:
The update in the fedora CI containers used for unit tests broke CI. Since the fix is only in develop, I decided to use commit on `develop` for the moment and require `Kokkos` >= 4.0.0.0 in general. Related to #88.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -33,9 +33,15 @@ jobs:
         run: |
           git submodule init
           git submodule update
-      - name: Install Kokkos
+     - name: Checkout Kokkos
+        uses: actions/checkout@v2.2.0
+        with:
+          repository: kokkos/kokkos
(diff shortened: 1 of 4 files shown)


dalg24 (approved):

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -33,9 +33,15 @@ jobs:
...
           git submodule update
-      - name: Install Kokkos
+     - name: Checkout Kokkos
+        uses: actions/checkout@v2.2.0
+        with:
+          repository: kokkos/kokkos
+          # FIXME Use 4.0.00 or later when released
+          ref: b5bd709fd615a0a5bc78492f171307bdafa2f53a
+          path: kokkos_src
+     - name: Install Kokkos
```
dalg24, line 43: Indentation is wrong which caused CI to not run

PR #93, opened 2023-01-22
Drop (unused) `Impl::eliminate_warning_for_lock_array` CUDA/HIP functions

dalg24:
Not able to re-open #89, sorry about the noise


masterleinad (approved):
Looks like https://github.com/kokkos/kokkos/pull/5786 doesn't show any warnings so this should be fine indeed.

crtrott (approved):
Hm ok. Checked the output from the various CUDA tests inside of Kokkos and didn't see any warning. So I guess its ok.

PR #94, opened 2023-01-30
Let device lock array ptrs be in constant memory regardless of whether RDC is enabled or not

dalg24:
Was in constant memory only when relocatable device code is enabled.

Corresponding to https://github.com/kokkos/kokkos/pull/5825


masterleinad (approved):
Looks good to me.

trws (approved):
Looks good to me.

crtrott (approved):

PR #100, opened 2023-03-16
Enforce that CUDA/HIP separable compilation configuration is consistent downstream

dalg24:
Corresponds to kokkos/kokkos#5991

Define `DESUL_ATOMICS_ENABLE_{CUDA,HIP}_SEPARABLE_COMPILATION` macros in the configurations and enforce that the relocatable device code mode is consistent to prevent ODR violations.
These are enabled using the `{CUDA,HIP}_SEPARABLE_COMPILATION` options and we export `desul_atomics_ENABLE_{CUDA,HIP}_SEPARABLE_COMPILATION` variables with CMake.


atomics/include/desul/atomics/Macros.hpp
```diff
@@ -11,6 +11,25 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Config.hpp>
 
+// Intercept incompatible relocatable device code mode which leads to ODR violations
+#ifdef DESUL_ATOMICS_ENABLE_CUDA
+#if (defined(DESUL_ATOMICS_ENABLE_CUDA_SEPARABLE_COMPILATION) &&  \
+     !defined(__CUDACC_RDC__)) ||                                 \
+    (!defined(DESUL_ATOMICS_ENABLE_CUDA_SEPARABLE_COMPILATION) && \
+     defined(__CUDACC_RDC__))
```
Rombur, line 19: CUDA Clang defines `__CLANG_RDC__ not `___CUDACC__RDC__` so we should probably  check for that here too.

dalg24: You're bringing up a good point although I don't think you can actually use relocatable device code with Clang+CUDA.  If I recall correctly, we'd need to link to NVCC to get it to work.
See https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/cmake/kokkos_enable_options.cmake#L150-L156

Rombur: You're right, Kokkos doesn't support RDC with Clang but I don't know about RAJA. Also we may want to add the check now so we don't forget later. Anyway, it's pretty minor. I am fine waiting for Kokkos to support RDC+clang to add the check.

trws: I'd probably say we add the check. The Kokkos support question comes down to build system config. It seems safer to make sure it works even if something downstream does use clang-CUDA with RDC.

dalg24: Commenting here for reference.  The `__CLANG_RDC__` macro was added in Clang 14 in
https://github.com/llvm/llvm-project/commit/d55f05d9f7dc8aa0468a9db19d84671acf05823a

dalg24: Done

trws (approved):

PR #104, opened 2023-03-20
Fix warning with NVC++

dalg24:
Fixup for #99
Getting warnings with NVC++ from NVHPC SDK 23.1
```
"<desul>/include/desul/atomics/Common.hpp", line 86: warning: integer conversion resulted in a change of sign [integer_sign_change]
    static constexpr uint32_t value = -1;
```
                                      ^


masterleinad (approved):

dalg24:
Turns out NVCC also raises warning https://godbolt.org/z/1c58TKfP7

trws (approved):

PR #107, opened 2023-05-31
Fix for OpenMP atomics with nvhpc.

rgayatri23:
The PR avoids the use of `memory-order-clause` with the `flush` construct when using OpenMP with NVHPC compiler.


dalg24 (approved):

dalg24:
What version of NVHPC did you test with?

rgayatri23:
> What version of NVHPC did you test with?

I have tested this with nvhpc/22.7 and nvhpc/23.1 (thats the latest as of now)

trws (approved):
_Sigh_ nvhpc... next we'll have to do it for the scope clauses as well.  Either way, thanks for the contribution!.

PR #108, opened 2023-09-13
Add support for OpenACC atomics

dalg24:
As far as the rest of the desul atomics are concerned the main change is we decorate all generic template function with the `DESUL_IMPL_ACC_ROUTINE_DIRECTIVE` macro.

Corresponds to ~~https://github.com/kokkos/kokkos/pull/5955~~ **edit** superseded by https://github.com/kokkos/kokkos/pull/6446


atomics/include/desul/atomics/Lock_Based_Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,81 @@
...
+          class MemoryScope,
+          // equivalent to:
+          //   requires !atomic_always_lock_free(sizeof(T))
+          std::enable_if_t<!atomic_always_lock_free(sizeof(T)), int> = 0>
+inline T device_atomic_fetch_oper(const Oper& op,
+                                  T* const dest,
+                                  dont_deduce_this_parameter_t<const T> val,
+                                  MemoryOrder /*order*/,
+                                  MemoryScope scope) {
+  if (acc_on_device(acc_device_not_host)) {
```
dalg24, line 32: @seyonglee I think this too should be printed unconditionally

seyonglee (approved):
It looks good to me.

atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,145 @@
...
+    // unlock_address_openacc((void*)dest, scope);
+    return return_val;
+  }
+}
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+T device_atomic_compare_exchange(
+    T* dest, T compare, T value, MemoryOrder, MemoryScope scope) {
+  if constexpr (std::is_integral_v<T> && ((sizeof(T) == 4) || (sizeof(T) == 8))) {
```
crtrott, line 55: why only integral here, and then still cast to some other integral below?

atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+namespace desul {
+namespace Impl {
+
+#ifdef __NVCOMPILER
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+T device_atomic_exchange(T* dest, T value, MemoryOrder, MemoryScope /*scope*/) {
+  if constexpr (std::is_arithmetic_v<T> && ((sizeof(T) == 4) || (sizeof(T) == 8))) {
```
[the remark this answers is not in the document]
dalg24: We were getting some weird ass compiler error with the conversion.  Will dig out the error message I think Seyong posted it somewhere.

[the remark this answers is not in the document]
seyonglee: NVHPC compiler complains if converting between integer and floating types during atomic operations.

atomics/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,427 @@
...
+}
+
+#ifdef __NVCOMPILER
+#pragma acc routine seq
+template <class T>
+std::enable_if_t<is_openacc_arithmetic_type_v<T>, T> device_atomic_max_fetch(
+    T* ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
+  tmp = atomicMax(ptr, val);
+  tmp = std::max(tmp, val);
```
crtrott, line 323: ?? signing twice why? Silencing warning for unused return argument?

atomics/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,427 @@
...
+#endif
+
+#ifdef __NVCOMPILER
+#pragma acc routine seq
+template <class T>
+std::enable_if_t<is_openacc_arithmetic_type_v<T>, T> device_atomic_min_fetch(
+    T* ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
+  tmp = atomicMin(ptr, val);
+  tmp = std::min(tmp, val);
```
crtrott, line 335: same question

atomics/include/desul/atomics/Thread_Fence_OpenACC.hpp
```diff
@@ -0,0 +1,25 @@
...
+#define DESUL_ATOMICS_THREAD_FENCE_OPENACC_HPP_
+
+namespace desul {
+namespace Impl {
+
+#pragma acc routine seq
+template <class MemoryOrder, class MemoryScope>
+void device_atomic_thread_fence(MemoryOrder, MemoryScope) {
+  // FIXME_OPENACC: The current OpenACC standard does not support explicit thread fence
+  // operations.
```
crtrott, line 19: you called atomicBlah in some other places, I assume that is essentially using CUDA intrinsics. Can't we do the same here?

crtrott (changes requested):
I don't get the atomic_exchange implementation choices.

atomics/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: The `atomic_max_fetch()` should return (fetch) the result of the max operation, but `atomicMax()` returns old value pointed by `ptr`. Therefore, we must again calculate `max(old-value, val)` to return it.

seyonglee: An alternative option will be to return `*ptr` instead of re-calculating it, but I'm not sure which will be better.

atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: NVHPC supports `atomicCAS()` only for specific types of integer and floating types and does not allow casting between integer types and floating types.

atomics/include/desul/atomics/Fetch_Op_OpenACC.hpp
dalg24: > An alternative option will be to return `*ptr` instead of re-calculating it, but I'm not sure which will be better.

No that would be a race condition.

Wha you do here is fine.

seyonglee: You are correct; I forgot why I did this way.

atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
dalg24: I just checked and atomicCAS does not like `int64_t` so we have to do that casting dance once way or another

dalg24: I added a comment to clarify why the code is that way

dalg24: Actually this seems to work
```diff
--- a/tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
+++ b/tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
@@ -23,7 +23,8 @@ namespace Impl {
 #pragma acc routine seq
 template <class T, class MemoryOrder, class MemoryScope>
 T device_atomic_exchange(T* dest, T value, MemoryOrder, MemoryScope /*scope*/) {
-  if constexpr (std::is_arithmetic_v<T> && ((sizeof(T) == 4) || (sizeof(T) == 8))) {
+
+  if constexpr (std::is_floating_point_v<T>) {
     T return_val;
 #pragma acc atomic capture
     {
@@ -31,6 +32,18 @@ T device_atomic_exchange(T* dest, T value, MemoryOrder, MemoryScope /*scope*/) {
       *dest = value;
     }
     return return_val;
+  } else if constexpr (std::is_arithmetic_v<T> && ((sizeof(T) == 4) || (sizeof(T) == 8))) {
+    static_assert(sizeof(unsigned int) == 4);
+    static_assert(sizeof(unsigned long long int) == 8);
+    using cas_t =
+        std::conditional_t<(sizeof(T) == 4), unsigned int, unsigned long long int>;
+    cas_t return_val;
+#pragma acc atomic capture
+    {
+      return_val = *reinterpret_cast<cas_t*>(dest);
+      *dest = value;
+    }
+    return reinterpret_cast<T&>(return_val);
   } else {
     // FIXME_OPENACC
     printf(
@@ -52,7 +65,13 @@ T device_atomic_exchange(T* dest, T value, MemoryOrder, MemoryScope /*scope*/) {
 template <class T, class MemoryOrder, class MemoryScope>
 T device_atomic_compare_exchange(
     T* dest, T compare, T value, MemoryOrder, MemoryScope scope) {
-  if constexpr (std::is_integral_v<T> && ((sizeof(T) == 4) || (sizeof(T) == 8))) {
+#ifdef DESUL_CUDA_ARCH_IS_PRE_PASCAL
+  if constexpr (std::is_same_v<T, float>) {
+#else
+  if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double>) {
+#endif
+    return atomicCAS(dest, compare, value);
+  } else if constexpr ((sizeof(T) == 4) || (sizeof(T) == 8)) {
     static_assert(sizeof(unsigned int) == 4);
     static_assert(sizeof(unsigned long long int) == 8);
     using cas_t =
@@ -61,12 +80,6 @@ T device_atomic_compare_exchange(
                                  reinterpret_cast<cas_t&>(compare),
                                  reinterpret_cast<cas_t&>(value));
     return reinterpret_cast<T&>(return_val);
-#ifdef DESUL_CUDA_ARCH_IS_PRE_PASCAL
-  } else if constexpr (std::is_same_v<T, float>) {
-#else
-  } else if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double>) {
-#endif
-    return atomicCAS(dest, compare, value);
   } else {
     // FIXME_OPENACC
     printf(

Comment (dalg24) [reply atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp:59]: But not if I drop the `is_arithmetic_v<T>` constraint in the `atomic_exchange`.
Getting an `Error: Detected unexpected atomic store opcode.`

Comment (seyonglee) [reply atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp:59]: Will the new version work for `long double` type? 
For example, if `sizeof(long double) == 8`, then `device_atomic_compare_exchange()` will fail due to casting between `unsigned long long int` and `long double`.

**Review by crtrott — DISMISSED** [review]

File: atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp [hunk 7]
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+T device_atomic_compare_exchange(
+    T* dest, T compare, T value, MemoryOrder, MemoryScope scope) {
+  // FIXME_OPENACC
+  printf(
+      "DESUL error in device_atomic_compare_exchange(): Not supported atomic operation "
+      "in the OpenACC backend\n");
```
trws, line 130: Given the issue with device_atomic_thread_fence being unimplementable, I get having a warning but why a printf, and why just this one rather than all of them when using a non-relaxed atomic?

dalg24: What else than printf could we do?  As far as I know we cannot abort from the device with OpenACC
https://github.com/kokkos/kokkos/blob/0e5aa150355cc5163198ae1cbba7d89fef2ab61f/core/src/Kokkos_Abort.hpp#L86-L87

This is the generic fallback so the warning will be raised if there is no overload that is more specialized.

seyonglee: The current OpenACC atomic implementation is tested only by the NVHPC compiler (and this PR includes implementations only for the NVHPC compiler), but the general Kokkos OpenACC backend is supposed to support both the NVHPC compiler and Clacc/LLVM compiler.
After this PR is merged, I will implement the specialized versions for the Clacc OpenACC compiler.
(Clacc OpenACC compiler is built on OpenMP, and OpenMP supports both atomic_compare_exchange() and atomic_thread_fence().)

trws: Ok, that works for me then.  I'll take a pass over it but IIRC this looked good the last time I checked.

crtrott (approved):

PR #111, opened 2023-10-13
(WIP) Add stack array class

adayton1:


src/desul/Macros.h
```diff
@@ -0,0 +1,16 @@
+#ifndef DESUL_MACROS_H
+#define DESUL_MACROS_H
+
+#if defined(__CUDACC__) || defined(__HIPCC__)
+#define DESUL_GPUCC
+#define DESUL_HOST_DEVICE __host__ __device__
+#else // defined(CARE_GPUCC)
+#define DESUL_HOST_DEVICE
+#endif // defined(CARE_GPUCC)
```
MrBurmark, line 9: CAREful

src/desul/array.hpp
```diff
@@ -0,0 +1,310 @@
...
+      DESUL_HOST_DEVICE constexpr const_iterator cend() const noexcept {
+         return &elements[size()];
+      }
+
+      DESUL_HOST_DEVICE constexpr bool empty() const noexcept {
+         return size() == 0;
+      }
+
+      DESUL_HOST_DEVICE constexpr size_type size() const noexcept {
+         return size();
```
MrBurmark, line 121:
```suggestion
         return N;
```

src/desul/array.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+      DESUL_HOST_DEVICE constexpr void fill(const T& value) {
+         for (std::size_t i = 0; i < size(); ++i) {
+            elements[i] = value;
+         }
+      }
+
+      DESUL_HOST_DEVICE constexpr void swap(array& other) noexcept(std::is_nothrow_swappable_v<T>) {
+         for (std::size_t i = 0; i < size(); ++i) {
+            swap(elements[i], other[i]);
```
MrBurmark, line 136: Add a `using std::swap`?

src/desul/array.hpp
```diff
@@ -0,0 +1,310 @@
...
+   template <class T>
+   inline constexpr std::size_t tuple_size_v = tuple_size<T>::value;
+
+   template <std::size_t I, class T>
+   struct tuple_element;
+    
+   template <std::size_t I, class T, std::size_t N>
+   struct tuple_element<I, array<T,N>> {
+      using type = T;
+   };
```
MrBurmark, line 303: This should be a specialization of std::tuple_element

src/desul/array.hpp
```diff
@@ -0,0 +1,310 @@
...
+   template <class T, std::size_t N>
+   DESUL_HOST_DEVICE constexpr array<std::remove_cv_t<T>, N> to_array(T (&&a)[N]) {
+      return detail::to_array_impl(std::move(a), make_index_sequence<N>{});
+   }
+#endif
+
+   template <class T, std::size_t N>
+   struct tuple_size<array<T, N>> :
+      std::integral_constant<std::size_t, N>
+   { };
```
MrBurmark, line 292: This should be a specializaiton of std::tuple_size

src/desul/utility.hpp
```diff
@@ -0,0 +1,46 @@
+#if !defined(DESUL_UTILITY_HPP)
+#define DESUL_UTILITY_HPP
+
+#include <cstddef>
+#include <type_traits>
+
+namespace desul {
+   template <class T, T... Ints>
+   struct integer_sequence {
```
MrBurmark, line 9: How much of this do we need to recreate? We can probably just bring most into desul via using.

src/desul/array.hpp
adayton1: Good catch. Thanks!

src/desul/Macros.h
adayton1: Fixed.

src/desul/array.hpp
adayton1: I added a desul::swap function and am calling that directly now.

src/desul/utility.hpp
adayton1: I think it depends on whether or not we want to require the --expt-relaxed-constexpr compiler flag with nvcc and if there is an equivalent to that for rocm/hip compilers.

src/desul/array.hpp
adayton1: Fixed.

src/desul/array.hpp
adayton1: Fixed.

src/desul/Macros.h
```diff
@@ -4,9 +4,9 @@
 #if defined(__CUDACC__) || defined(__HIPCC__)
```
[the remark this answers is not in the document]
adayton1: I've added what I think will work for SYCL. I'll need someone else with expertise in OpenMP target offload and OpenACC to figure out a good pattern for those programming models.

src/desul/utility.hpp
adayton1: But anything that does not have explicitly defined functions (like most type traits), I'm definitely supportive of bringing them in with using statements.

trws: I see it in the comments, but if you need integer_sequence stuff, take that wholesale from camp or somewhere.  It's a serious pain to get `make_integer_sequence` to perform well portably across compilers, and to cover some of the corner cases, no reason to go through that pain again.

src/desul/Macros.h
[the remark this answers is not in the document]
trws: I think all that logic exists in the atomics subfolder, perhaps we can promote it, or if we need something additional we have a pretty complete cascade for. ompiler and language detection in camp: https://github.com/trws/camp/blob/fix_ci/include/camp/defines.hpp

src/desul/utility.hpp
adayton1: I'm going to leave it out for now and bring it in from camp in a future merge request.

src/desul/Macros.h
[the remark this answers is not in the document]
adayton1: So on that note, it seems as if the atomics were really designed as a standalone library rather than as part of a set of shared components. What is the overall layout we want for this library if we start adding more components? I can see two potential approaches:

1. The current approach: have a main folder for each grouping of components (atomics, containers, math, algorithms, etc...). The main difficulty is deciding how to group components and dealing with components that reach across groups.
2. Have a <repo root>/src/desul directory that contains the public headers, and then internal implementation details are put in subdirectories in that main directory as needed.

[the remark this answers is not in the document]
trws: My original hope was that we would go for something somewhat between the two.

* Each main folder contains a component
* We'll want to factor out a "common" component for things like determining the compiler and so-forth I think as well so we don't need to reach between them for unrelated things
* At the top level, have a build that builds all the components, and offers a desul include directory with a `desul.hpp` that includes all the components.  While the mega-header approach isn't very nice for headers, it's _very_ nice when we start talking about modules, so planning ahead just a bit.

src/desul/array.hpp
```diff
@@ -0,0 +1,301 @@
...
+         else {
+            return false;
+         }
+      }
+
+      return false;
+   }
+
+   template <class T, std::size_t N>
+   DESUL_HOST_DEVICE constexpr bool operator<=(const array<T, N>& lhs,
```
MrBurmark, line 176: Should this be implemented in terms of operator<?

src/desul/array.hpp
```diff
@@ -0,0 +1,301 @@
...
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] < rhs[i]) {
+            return true;
+         }
+         else if (lhs[i] == rhs[i]) {
+            continue;
+         }
+         else {
+            return false;
+         }
```
MrBurmark, line 169: Should this all be done in terms of operator<?
```suggestion
         else if ( rhs[i] < lhs[i]) {
            return false;
         }
```

src/desul/array.hpp
```diff
@@ -0,0 +1,301 @@
...
+      }
+
+      value_type elements[N];
+   };
+
+   template <class T, std::size_t N>
+   DESUL_HOST_DEVICE constexpr bool operator==(const array<T, N>& lhs,
+                                               const array<T, N>& rhs) {
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] != rhs[i]) {
```
MrBurmark, line 143: Should we only use operator==?

src/desul/array.hpp
```diff
@@ -0,0 +1,301 @@
...
+      using const_pointer = const value_type*;
+      using iterator = pointer;
+      using const_iterator = const_pointer;
+      using reverse_iterator = pointer;
+      using const_reverse_iterator = const_pointer;
+
+      constexpr reference at(size_type pos) {
+         if (pos >= size()) {
+            throw std::out_of_range{"desul::array::at detected out of range access"};
+         }
```
MrBurmark, line 39: Should we think about using assert in device code instead of throwing as a away to check bounds?

PR #114, opened 2023-10-24
SYCL: Implement DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION path

masterleinad:
Follow-up to https://github.com/desul/desul/pull/113 and https://github.com/kokkos/kokkos/pull/6531#discussion_r1366248056. Based on https://github.com/desul/desul/pull/113.
With `SYCL_EXT_ONEAPI_DEVICE_GLOBAL` being defined, we can also implement the `DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION` path analogously to the `Cuda` and `HIP` implementation. This now also seems to work fine when using shared libraries.

Corresponds to https://github.com/kokkos/kokkos/pull/6534.


atomics/src/Lock_Array_SYCL.cpp
```diff
@@ -63,7 +53,10 @@ void finalize_lock_arrays_sycl<int>(sycl::queue q) {
   sycl::free(SYCL_SPACE_ATOMIC_LOCKS_NODE_h, q);
   SYCL_SPACE_ATOMIC_LOCKS_DEVICE_h = nullptr;
   SYCL_SPACE_ATOMIC_LOCKS_NODE_h = nullptr;
+#ifdef DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION
```
dalg24, line 56: I see also guarded it for CUDA and HIP but it is not clear to me why we did.
Not asking you to change but do you know why that would be needed?

masterleinad: My best guess is that we do it for the same reason as we assign `nullptr` to `[CUDA|HIP]_SPACE_ATOMIC_LOCKS_[DEVICE|NODE]_h` and that that is to avoid dangling pointers to freed allocations in case we don't safeguard probably against executing kernels after this function has been called. This is very similar to all in the `nullptr` assignments in `ExecutionSpace::finalize`.

atomics/include/desul/atomics/Lock_Array_SYCL.hpp
```diff
@@ -62,9 +62,21 @@ void finalize_lock_arrays_sycl(sycl::queue q);
  * declaration here must be extern). This one instance will be initialized
  * by initialize_host_sycl_lock_arrays and need not be modified afterwards.
  */
-SYCL_EXTERNAL extern sycl_device_global<int32_t*> SYCL_SPACE_ATOMIC_LOCKS_DEVICE;
+#ifdef DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION
+SYCL_EXTERNAL extern
+#else
+static
```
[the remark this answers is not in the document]
masterleinad: Otherwise, we would have duplicated symbols violating ODR. IIRC, `__device__ __contant__` global variables are implicitly `static` for `CUDA` and `HIP`, also see https://github.com/desul/desul/blob/0a0dfd3a4b53030913f970f0a28714a0415e147a/atomics/include/desul/atomics/Lock_Array_CUDA.hpp#L63-L66 (which means I have to update the comment here as well).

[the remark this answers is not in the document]
masterleinad: 2c192da

atomics/include/desul/atomics/Lock_Array_SYCL.hpp
```diff
@@ -57,14 +57,35 @@ void finalize_lock_arrays_sycl(sycl::queue q);
...
+ * instance will be initialized by initialize_host_sycl_lock_arrays and need not be
+ * modified afterwards.
+ *
+ * When relocatable device code is disabled, an instance of this variable will be
+ * created in every translation unit that sees this header file (we make this clear by
+ * marking it static, meaning no other translation unit can link to it). Since the
+ * Kokkos_SYCL_Locks.cpp translation unit cannot initialize the instances in other
+ * translation units, we must update this SYCL global variable based on the Host global
+ * variable prior to running any kernels that will use it. That is the purpose of the
+ * ensure_cuda_lock_arrays_on_device function.
```
dalg24, line 72:
```suggestion
 * ensure_sycl_lock_arrays_on_device function.
```

dalg24 (approved):

atomics/src/Lock_Array_SYCL.cpp
masterleinad: Yeah, that's interesting.

With RDC, we have only one device variable that is the same for all translation units. Assuming that, `init_lock_arrays_*` was called, `copy_*_lock_arrays_to_device` won't do anything since the copy is only executed once. Otherwise, we actually copy the `nullptr`.
Without RDC, we have one device variable per translation unit that we are setting up via `ensure_*_lock_arrays_on_device` which basically calls `copy_*_lock_arrays_to_device` again. In this case, we can't reset all the device variables anyway.

I think it makes the most sense to either remove the guard here or add it in `init_lock_arrays_*` as well. If we assume that `init_lock_arrays_*` is called before `finalize_lock_arrays_*` we can also remove the call here. I'm happy to open a follow-up pull request that discusses/addresses that uniformly across the backends.

trws: That sounds like a reasonable approach, but I think in practice we can end up in a situation with more than one lock array _even with RDC turned on_ because of the split between separate dynamic libraries.  I sincerely hope nobody's doing that, but we're only guaranteed to have a single array _per dynamic library_ not globally.

dalg24 (approved):

atomics/src/Lock_Array_SYCL.cpp
dalg24: Let's close this parenthesis and address this in a follow up.
In the meantime I think this is ready.

trws: I agree, looks really good actually.

trws (approved):

PR #116, opened 2023-10-27
fix cuda target name

trws:
This is one of the dumber mistakes I've made in a while, messed up the blt target name for an updated one that doesn't work yet. This fixes it.


dalg24 (changes requested):
This is a partially undoing #115

Please elaborate why we should not be using the qualified target and why the HIP one does not need the same fix.

trws:
Of course, the cause is that the qualified target doesn’t work for Cuda yet.  I could add an alias to hide the issue until the blt update comes down, or leave it like this.  The hip one, ironically, is both already there and the original bare hip name was removed, hence my confusion and this whole mess.

dalg24:
What about something like
```cmake
# FIXME_BLT
if(TARGET cuda AND NOT TARGET blt::cuda)
  add_library(blt::cuda ALIAS cuda)
endif()

dalg24 (approved):

masterleinad (approved):

PR #117, opened 2023-11-01
Desul config control

trws:
This adds some new `DESUL_*` defines to allow turning off something that is otherwise `ENABLE_*`ed.  This should have no impact on Kokkos usage, but makes it easier to allow a project with say `ENABLE_OPENMP` on to turn it off specifically for desul.  This is also essentially the same setup that RAJA, camp, etc. have.  It also cleans up blt loading, and refers to a BLT PR to fix the `blt::cuda` target to work properly upstream (our alias doesn't work properly in some cases).

https://github.com/LLNL/blt/pull/664


rchen20 (approved):
This looks good to me as is, when I test it with RAJA.

CMakeLists.txt
```diff
@@ -9,25 +9,25 @@ if (NOT BLT_CXX_STD)
...
+      if (NOT ${option})
+         message(FATAL_ERROR "DESUL_${option} set to On, but ${option} is Off. Please set ${option} to On enable this feature.")
+      endif ()
+   endif ()
+endforeach ()
+
+
+include(CMakeDependentOption)
+option(ENABLE_SYCL "Enable SYCL support" OFF) # option does not actually yet exist on the BLT side
+cmake_dependent_option(DESUL_ENABLE_CUDA "Build with CUDA support" On "ENABLE_CUDA" Off)
```
dalg24, line 24: I am not a big fan of mixing the spelling `ON/OFF` (line above) and "On/Off"
I would prefer if we stuck to uppercase.

trws: Sure, no problem.

trws:
Ok, I think this is cleaned up and ready to go.

dalg24 (approved):

PR #124, opened 2024-03-12
Revisit class providing atomic operations for non-atomic objects

dalg24:
```C++
template <typename T, typename MemoryOrder, typename MemoryScope>
struct scoped_atomic_ref;
```
The `desul::scoped_atomic_ref` was intended as an extension of the C++20 `std::atomic_ref`.
One notable change was the `MemoryOrder` template parameter that was used to modify the default order of all atomic operations.

The class template had never been used in production and we recently discovered it was broken (see #123).

We have been prototyping an implementation of a desul-based atomic accessor for `mdspan` in Kokkos and we realized that the design of `desul::scoped_atomic_ref` is not what we need.  We want to be able to use the atomic accessor in `Kokkos::View` when the atomic memory traits is specified.  It requires for the accessor to return a proxy type that supports all arithmetic operators that some potentially user-defined scalar type would provide.

This PR suggests to drop `scoped_atomic_ref` and replace it with `desul::AtomicRef` with a new design that can satisfy what Kokkos need and that we think is more in-line with the spirit of the generic atomic free functions that are provided by desul.  The main idea is that `AtomicRef` is just an alternate syntax to get to the atomics operations and that the same constraints and preconditions apply (we should seriously think about discussing these in the near future but essentially we suggest to make `AtomicRef`just behave the same as the free functions).

The current implementation is tested in kokkos/kokkos#6872

Open questions:
* should we only provide `fetch_op` member functions or do we also want `op_fetch`?
* do we all agree that trailing (templated) memory order and/or memory scope arguments in member functions are silly and we don't want them?
   ```C++
   // Alternative that I don't like
   template <typename T, typename DefaultMemoryOrder, typename DefaultMemoryScope>
   class AtomicRef {
     template <typename MemoryOrder = DefaultMemoryOrder, typename MemoryScope = DefaultMemoryScope>
     DESUL_FUNCTION T fetch_op(T val, MemoryOrder = MemoryOrder(), MemoryScope = MemoryScope()) const noexcept;
   };
   // VS what is currently proposed
   template <typename T, typename MemoryOrder, typename MemoryScope>
   class AtomicRef {
     DESUL_FUNCTION T fetch_op(T val) const noexcept;
   };
   ```
* should we consider adding an `AtomicRef` specialization for pointers with corresponding overloads of the free functions?

### DIFF
--- a/atomics/include/desul/atomics/Atomic_Ref.hpp
+++ b/atomics/include/desul/atomics/Atomic_Ref.hpp
@@ -6,60 +6,21 @@ Source: https://github.com/desul/desul
 SPDX-License-Identifier: (BSD-3-Clause)
 */
 
-#ifndef DESUL_ATOMIC_REF_IMPL_HPP_
-#define DESUL_ATOMIC_REF_IMPL_HPP_
+#ifndef DESUL_ATOMIC_REF_HPP_
+#define DESUL_ATOMIC_REF_HPP_
 
-#include <cstddef>
 #include <desul/atomics/Common.hpp>
 #include <desul/atomics/Generic.hpp>
 #include <desul/atomics/Macros.hpp>
-#include <memory>
-#include <type_traits>
 
 namespace desul {
-namespace Impl {
 
-// TODO current implementation is missing the following:
-// * member functions
-//   * wait
-//   * notify_one
-//   * notify_all
-
-template <typename T,
-          typename MemoryOrder,
-          typename MemoryScope,
-          bool = std::is_integral<T>{},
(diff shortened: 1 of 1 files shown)


trws (changes requested):
I like this direction.  After discussing with @dalg24 just now, we decided to add at least the exchange op to this and to go forward with it in general.  The alignment and always lock free setup is... really not safe, and is going to be left out pending both a use for it and a way to more reliably give a useful answer on a given platform.

crtrott (approved):

trws (approved):

PR #127, opened 2024-05-07
Use builtin atomic for compare_exchange

Rombur:
This PR  reworks atomic_compare for the HIP backend. It uses the builtin functions in order to use the MemoryOrder. This simplifies the current code and aligns it with `Compare_Exchange_GCC.hpp`

Link to the Kokkos PR: https://github.com/kokkos/kokkos/pull/7000

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
@@ -51,14 +51,6 @@ device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope) {
 }
 
 template <class T, class MemoryOrder, class MemoryScope>
-__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
-device_atomic_exchange(T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
-  T return_val = device_atomic_compare_exchange(
-      dest, compare, value, MemoryOrder(), MemoryScope());
-  return reinterpret_cast<T&>(return_val);
-}
-
-template <class T, class MemoryOrder, class MemoryScope>
 __device__ std::enable_if_t<!atomic_exchange_available_hip<T>::value, T>
 device_atomic_compare_exchange(
     T* const dest, T compare, T value, MemoryOrder, MemoryScope scope) {
--- a/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
@@ -40,35 +40,22 @@ device_atomic_compare_exchange(
   return compare;
 }
 
-template <class T, class MemoryScope>
+template <class T, class MemoryOrder, class MemoryScope>
 __device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
-device_atomic_exchange(T* const dest, T value, MemoryOrderRelaxed, MemoryScope) {
+device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope) {
   T return_val = __hip_atomic_exchange(dest,
                                        value,
-                                       HIPMemoryOrder<MemoryOrderRelaxed>::value,
+                                       HIPMemoryOrder<MemoryOrder>::value,
                                        HIPMemoryScope<MemoryScope>::value);
   return return_val;
 }
 
-template <class T, class MemoryScope>
+template <class T, class MemoryOrder, class MemoryScope>
 __device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
-device_atomic_exchange(
-    T* const dest, T compare, T value, MemoryOrderRelease, MemoryScope) {
+device_atomic_exchange(T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
   T return_val = device_atomic_compare_exchange(
-      dest, compare, value, MemoryOrderRelease(), MemoryScope());
+      dest, compare, value, MemoryOrder(), MemoryScope());
   return reinterpret_cast<T&>(return_val);
-}
-
-template <class T, class MemoryScope>
-__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
-device_atomic_exchange(
-    T* const dest, T /*compare*/, T value, MemoryOrderAcquire, MemoryScope) {
-  T return_val = __hip_atomic_exchange(dest,
-                                       value,
-                                       HIPMemoryOrder<MemoryOrderAcquire>::value,
-                                       HIPMemoryScope<MemoryScope>::value);
-  device_atomic_exchange(dest, value, MemoryOrderRelaxed(), MemoryScope());
-  return return_val;
 }
 
 template <class T, class MemoryOrder, class MemoryScope>
--- a/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
@@ -9,6 +9,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #ifndef DESUL_ATOMICS_COMPARE_EXCHANGE_HIP_HPP_
 #define DESUL_ATOMICS_COMPARE_EXCHANGE_HIP_HPP_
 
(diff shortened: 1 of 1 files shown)


atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
```diff
@@ -17,130 +18,72 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
-  static_assert(sizeof(unsigned long long int) == 8,
-                "this function assumes an unsigned long long is 64-bit");
-  unsigned long long int return_val =
-      atomicExch(reinterpret_cast<unsigned long long int*>(dest),
-                 reinterpret_cast<unsigned long long int&>(value));
-  return reinterpret_cast<T&>(return_val);
+  T return_val = __hip_atomic_exchange(dest, value,
+                                       HIPMemoryOrder<MemoryOrderRelaxed>::value,
+                                       HIPMemoryScope<MemoryScope>::value);
+  return return_val;
```
dalg24, line 46:
```suggestion
  return __hip_atomic_exchange(dest, value,
                               HIPMemoryOrder<MemoryOrderRelaxed>::value,
                               HIPMemoryScope<MemoryScope>::value);
```
(haven't checked the format)

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
```diff
@@ -17,130 +18,72 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
+  T return_val = __hip_atomic_exchange(dest, value,
+                                       HIPMemoryOrder<MemoryOrderAcqRel>::value,
+                                       HIPMemoryScope<MemoryScope>::value);
+  return return_val;
 }
 
 template <class T, class MemoryScope>
-__device__ std::enable_if_t<sizeof(T) == 4 || sizeof(T) == 8, T> device_atomic_exchange(
+__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T> device_atomic_exchange(
     T* const dest, T value, MemoryOrderSeqCst, MemoryScope) {
```
dalg24, line 78: Why would we provide this more specialized version?
Is it not covered by the generic overload L41?

dalg24: Did you mean to have all these `device_atomic_exchange` overload L49-L83 
https://github.com/desul/desul/blob/84cfa75787c2e2deafaa801fa3a45f8ad73c2c7a/atomics/include/desul/atomics/Compare_Exchange_HIP.hpp#L49-L83
or did you forget to remove?

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
```diff
@@ -17,130 +18,61 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
-      atomicCAS(reinterpret_cast<unsigned long long int*>(dest),
-                reinterpret_cast<unsigned long long int&>(compare),
-                reinterpret_cast<unsigned long long int&>(value));
-  return reinterpret_cast<T&>(return_val);
-}
+template <class T>
+struct atomic_exchange_available_hip {
+  constexpr static bool value =
+  ((sizeof(T) == 1 && alignof(T) == 1) || (sizeof(T) == 4 && alignof(T) == 4) ||
+   (sizeof(T) == 8 && alignof(T) == 8)) &&
```
Rombur, line 25: According the documentation, we should be able to use the builtin functions for sizeof = 1, 4, 8, and 16 but I got a warning with 16, so I dropped it.

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
```diff
@@ -17,130 +18,61 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
 template <class T, class MemoryScope>
-__device__ std::enable_if_t<sizeof(T) == 4 || sizeof(T) == 8, T>
-device_atomic_compare_exchange(
-    T* const dest, T compare, T value, MemoryOrderAcqRel, MemoryScope) {
-  atomic_thread_fence(MemoryOrderAcquire(), MemoryScope());
-  T return_val = atomic_compare_exchange(
-      dest, compare, value, MemoryOrderRelaxed(), MemoryScope());
-  atomic_thread_fence(MemoryOrderRelease(), MemoryScope());
+__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
+device_atomic_exchange(T* const dest, T value, MemoryOrderRelaxed, MemoryScope) {
```
dalg24, line 45: Did you mean to keep that overload?

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
```diff
@@ -17,130 +18,61 @@ SPDX-License-Identifier: (BSD-3-Clause)
...
-      atomicExch(reinterpret_cast<unsigned long long int*>(dest),
-                 reinterpret_cast<unsigned long long int&>(value));
-  return reinterpret_cast<T&>(return_val);
-}
-
-template <class T, class MemoryScope>
-__device__ std::enable_if_t<sizeof(T) == 4 || sizeof(T) == 8, T> device_atomic_exchange(
+__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
+device_atomic_exchange(
     T* const dest, T compare, T value, MemoryOrderRelease, MemoryScope) {
```
dalg24, line 56: Same question here.

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
Rombur: yes

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
Rombur: yes

atomics/include/desul/atomics/Compare_Exchange_HIP.hpp
dalg24: Please explain why we need them and the more generic one is not enough.

Rombur: Because the signature is different. The one you call more "generic" take a `compare` operator. This function doesn't.

dalg24: Why are we not doing
```C++
__device__ std::enable_if_t<atomic_exchange_available_hip<T>::value, T>
device_atomic_exchange(T* const dest, T value, MemoryOrderRelaxed, MemoryScope) {
  return __hip_atomic_exchange(dest,
                               value,
                               HIPMemoryOrder<MemoryOrder>::value,
                               HIPMemoryScope<MemoryScope>::value);
}

Rombur: We only support this API for a subset of the `MemoryOrder`. We could extent it but this PR just replace the implementation of the current API.

dalg24: > We only support this API for a subset of the `MemoryOrder`.

You mean the current overload set (before this PR) would not find a match for `MemoryOrderSeqCst`?

Please prefer the generic overload.  We'd probably want to handle constraints/mandates consistently.  We don;t need to handle it in this PR.  With the current proposed version, if trying to call seq_cst with a non-supported type that goes though the lock-based implementation would fail to compile with a possibly confusing error message.

Rombur: Sure, I'll do it. I just didn't want to add new functionality in a PR that was only supposed to change the implementation.

Rombur: I've added the generic implementation of `device_atomic_exchange`. I've also removed the `device_atomic_exchange` function that was taking an unused `compare` operator. This was only implemented for `MemoryOrderAcquire`. This function does not exist for the other backend and doesn't make sense too me. Why give a `compare` operator that is ignored when calling `device_atomic_exchange` when there is `device_atomic_compare_exchange` which uses the operator. Let me know if you want me to put this function back.

dalg24 (approved):

masterleinad (approved):

publixsubfan (approved):

PR #129, opened 2024-06-29
Add bitwise shift operations to AtomicRef

crtrott:
Doesn't look like we have unit tests yet at all for AtomicRef??

### DIFF
--- a/atomics/include/desul/atomics/Atomic_Ref.hpp
+++ b/atomics/include/desul/atomics/Atomic_Ref.hpp
@@ -95,6 +95,22 @@ class AtomicRef {
   DESUL_IMPL_DEFINE_ATOMIC_INCREMENT_DECREMENT(--, dec)
 
 #undef DESUL_IMPL_DEFINE_ATOMIC_INCREMENT_DECREMENT
+
+#define DESUL_IMPL_DEFINE_ATOMIC_BITWISE_SHIFT(COMPD_ASGMT, SHFT)                 \
+  DESUL_FUNCTION T fetch_##SHFT(unsigned int arg) const noexcept {                \
+    return desul::atomic_fetch_##SHFT(ptr_, arg, MemoryOrder(), MemoryScope());   \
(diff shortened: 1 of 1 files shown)


crtrott:
Note: I tested this version in Kokkos in my attempt to replace the guts of View with mdspan. That is how I found these ops missing compared to what our atomic reference object did in Kokkos.

atomics/include/desul/atomics/Atomic_Ref.hpp
```diff
@@ -79,6 +79,10 @@ class AtomicRef {
   DESUL_IMPL_DEFINE_ATOMIC_FETCH_OP(xor)
   DESUL_IMPL_DEFINE_ATOMIC_COMPOUND_ASSIGNMENT_OP(^=, xor)
   DESUL_IMPL_DEFINE_ATOMIC_FETCH_OP(nand)
```
dalg24, line 81: I think we should have them take the operand as unsigned int instead of T

dalg24:
> Doesn't look like we have unit tests yet at all for AtomicRef?? 

I had forced all kokkos atomics to go through atomic reference. But yes you're correct.

atomics/include/desul/atomics/Atomic_Ref.hpp
crtrott: isn't that true for mod too?

dalg24: I am talking about consistency with the free function signatures
https://github.com/desul/desul/blob/60c1115af231f4bfe0d70a3e0cf54b2cc0ff1594/atomics/include/desul/atomics/Fetch_Op_Generic.hpp#L57-L60
We currently define `mod` as taking the operand as a `T`.

dalg24:
The diff rendering was fishy, so I reset on the latest version, reapplied the changes I suggested, I push forced.

trws (approved):
Works for me.

PR #135, opened 2025-03-02
Add 128 bit atomics to overload set for SM_90

### LINKED ISSUE #133: cuda: hardware support for 128 bits data types is available starting from compute capability `9.x`
According to [this table](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#features-and-technical-specifications):
> Atomic functions operating on 128-bit integer values in global/shared memory

is supported from compute capability `9.x` and above. This means that data types like `complex<double>` are supported.

This should be carefully tested, but my guess is that hardware supported ops are faster than the lock-based ops from `desul`.

JBludau:
fixes #133

The [documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#features-and-technical-specification) lists 128bit atomic operations as supported for sm>=9.x.

These can be used for e.g. `complex<double>` (already covered in unit tests)

Corresponding Kokkos PR https://github.com/kokkos/kokkos/pull/8025

### DIFF
--- a/atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
+++ b/atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
@@ -1,0 +1,25 @@
+
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_EXCHANGE() \
+template<class ctype> \
+inline __device__ typename ::std::enable_if<sizeof(ctype)==16, ctype>::type device_atomic_exchange(ctype* dest, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  __int128_t asm_value = reinterpret_cast<__int128_t&>(value); \
+  __int128_t asm_result = 0u; \
+  asm volatile("atom.exch" __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER_ASM __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE_ASM ".b128" " %0,[%1],%2;" : "=q"(asm_result) : "l"(dest),"q"(asm_value) : "memory"); \
+  return reinterpret_cast<ctype&>(asm_result); \
+}
+
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_COMPARE_EXCHANGE() \
+template<class ctype> \
+inline __device__ typename ::std::enable_if<sizeof(ctype)==16, ctype>::type device_atomic_compare_exchange(ctype* dest, ctype compare, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  __int128_t asm_value = reinterpret_cast<__int128_t&>(value); \
+  __int128_t asm_compare = reinterpret_cast<__int128_t&>(compare); \
+  __int128_t asm_result = 0u; \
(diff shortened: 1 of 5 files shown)


dalg24:
Do you have a corresponding PR on the Kokkos side?
Did you run the benchmarks?  Please post the results.

dalg24 (commented):
I believe that we also need to enable lock-free operations for when the memory order is not relaxed, where IIUC we would now acquire a lock because we explicitly check for 32- or 64-bit types.
See `Compare_Exchange_CUDA.hpp`.
In `Lock_Based_Fetch_Op_CUDA.hpp` it looks like we use `atomic_always_lock_free` in our constraint.

atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
```diff
@@ -0,0 +1,25 @@
...
+  return reinterpret_cast<ctype&>(asm_result); \
+}
+
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_COMPARE_EXCHANGE() \
+template<class ctype> \
+inline __device__ typename ::std::enable_if<sizeof(ctype)==16, ctype>::type device_atomic_compare_exchange(ctype* dest, ctype compare, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  __int128_t asm_value = reinterpret_cast<__int128_t&>(value); \
+  __int128_t asm_compare = reinterpret_cast<__int128_t&>(compare); \
+  __int128_t asm_result = 0u; \
+  asm volatile("atom.cas" __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER_ASM __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE_ASM ".b128" " %0,[%1],%2,%3;" : "=q"(asm_result) : "l"(dest),"q"(asm_compare),"q"(asm_value) : "memory"); \
```
romintomasetti, line 17: Out of curiosity, why do you implement this in *asm* instead of using https://docs.nvidia.com/cuda/cuda-c-programming-guide/#atomiccas ?

atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
```diff
@@ -0,0 +1,25 @@
+
+#define __DESUL_IMPL_CUDA_ASM_ATOMIC_EXCHANGE() \
+template<class ctype> \
+inline __device__ typename ::std::enable_if<sizeof(ctype)==16, ctype>::type device_atomic_exchange(ctype* dest, ctype value, __DESUL_IMPL_CUDA_ASM_MEMORY_ORDER, __DESUL_IMPL_CUDA_ASM_MEMORY_SCOPE) { \
+  __int128_t asm_value = reinterpret_cast<__int128_t&>(value); \
```
romintomasetti, line 5: Have you seen https://docs.nvidia.com/cuda/cuda-c-programming-guide/#nv-atomic-exchange ?

atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
JBludau: I will check again why it is implemented in asm for Volta.

JBludau: So we opted for not continuing with the ptx since we can use one impl until we bump the CUDA version to 12.8

atomics/include/desul/atomics/cuda/cuda_cc9_asm_exchange_op.inc
JBludau: same for this

JBludau:
> Do you have a corresponding PR on the Kokkos side? Did you run the benchmarks? Please post the results.
Kokkos PR: https://github.com/kokkos/kokkos/pull/8025
Benchmarks: https://github.com/kokkos/kokkos/pull/8025#issuecomment-3313722654

dalg24 (approved):

PR #137, opened 2025-04-21
HIP: Implement atomic_op_fetch() in terms of atomic_fetch_op()

### LINKED ISSUE #136: Slow HIP atomic via AtomicRef += operator
In `AtomicRef`, operator `+=` is implemented in `Atomic_Ref.hpp` as 
```
DESUL_IMPL_DEFINE_ATOMIC_COMPOUND_ASSIGNMENT_OP(+=, add_fetch)

 ->

DESUL_FUNCTION T OP_FETCH(T arg) const noexcept {                           \
  return desul::atomic_add_fetch(ptr_, arg, MemoryOrder(), MemoryScope()); \
}
```
Since there is no "OP_FETCH" operation implemented for HIP, but only "FETCH_OP": 
```
#define DESUL_IMPL_HIP_ATOMIC_FETCH_OP(OP, T)                           \
  template <class MemoryOrder, class MemoryScope>                       \
  __device__ inline T device_atomic_fetch_##OP(                         \
      T* ptr, T val, MemoryOrder, MemoryScope) {                        \

    return __hip_atomic_fetch_##OP(ptr,                                 \
                                   val,                                 \
                                   HIPMemoryOrder<MemoryOrder>::value,  \
                                   HIPMemoryScope<MemoryScope>::value); \
  }
```
the generic versions get called, which are much slower.

dalg24:
Compute atomic_op_fetch(ptr, arg) as the op applied to atomic_fetch_op(ptr, arg) and arg

Fix #136 
Corresponds to kokkos/kokkos#8010

### DIFF
--- a/atomics/include/desul/atomics/Fetch_Op_HIP.hpp
+++ b/atomics/include/desul/atomics/Fetch_Op_HIP.hpp
@@ -10,6 +10,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_FECH_OP_HIP_HPP_
 
 #include <desul/atomics/Adapt_HIP.hpp>
+#include <desul/atomics/Operator_Function_Objects.hpp>
 
 namespace desul {
 namespace Impl {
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
We should also look at the other backends. Using the operator repeatedly is likely always better if there is a native `atomic_OP_fetch` implementation (instead of going through `compare_exchange`).

dalg24:
> We should also look at the other backends. Using the operator repeatedly is likely always better if there is a native `atomic_OP_fetch` implementation (instead of going through `compare_exchange`).

Agreed, but not as part of this PR.
Would you please handle SYCL?

crtrott (changes requested):
Unless someone really demonstrates a performance benefit: just lets implement the generic versions of the op_fetch atomics in terms of calling fetch_op atomics by hitting the public overloads and thus get the proper specialized ones from the backend. This way we implement it once and have the thing addressed for everyone. I mean as far as I can tell neither CUDA, HIP nor SYLC provide op_fetch anyway. So doing it one go for everyone is referable in my mind.

dalg24:
Closing in favor of #139

PR #143, opened 2025-08-18
Enable tests in desul

masterleinad:
It seems, the CI doesn't enable the tests. https://github.com/desul/desul/actions/runs/15192835955/job/42729728895 for example only prints
```
Test project /__w/desul/desul/build
    Start 1: blt_gtest_smoke
1/1 Test #1: blt_gtest_smoke ..................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.00 sec
```

PR #144, opened 2025-08-18
Run CI with -Werror for -Wall and -Wextra

masterleinad:


dalg24 (commented):
Blocked on #143

masterleinad:
> Blocked on #143

Yes, that would be the preferred order.

PR #146, opened 2025-08-18
Fix building with MSVC+Cuda

masterleinad:
Extracts the relevant changes for allowing compiling on Windows with shared libraries from kokkos/kokkos#8324. Testing (using `Kokkos`) is difficult before merging a respective pull request in `Kokkos`.

Since we don't want to use `desul`'s `CMake` configuration when bundling with `Kokkos`, using `generate_export_header` would be cumbersome (instead of extracting relevant pieces as done here).

### DIFF
--- a/atomics/CMakeLists.txt
+++ b/atomics/CMakeLists.txt
@@ -40,7 +40,7 @@ if(DESUL_ENABLE_OPENMP)
   set(DESUL_ATOMICS_ENABLE_OPENMP ON)
 endif()
 if (BUILD_SHARED_LIBS)
-  set(DESUL_BUILD_SHARED_LIBS ON)
+  set(DESUL_IMPL_BUILD_SHARED_LIBS ON)
 endif()
 configure_file(
   ${PROJECT_SOURCE_DIR}/atomics/Config.hpp.cmake.in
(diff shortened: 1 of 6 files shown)


atomics/CMakeLists.txt
```diff
@@ -57,6 +60,9 @@ target_include_directories (desul_atomics PUBLIC
   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>
   )
+if (BUILD_SHARED_LIBS)
+  target_compile_definitions(desul_atomics PRIVATE DESUL_EXPORT_SYMBOLS)
```
dalg24, line 64: Why do we need both `DESUL_BUILD_SHARED_LIBS` and `DESUL_EXPORT_SYMBOLS`?
Why do we put the first one in the config file and the second as a compile definition?

masterleinad: We only need to export/import anything when building a shared library otherwise we just create overhead.
We need to export when building the library and import when using it. Thus, we can't control the behavior using a config file but must rather use build flags (that aren't exported).

trws: What overhead? Is there a downside to just always applying export or import when in win32?

JBludau (approved):
lgtm

trws (dismissed):

masterleinad:
@crtrott Any opinions here? Do you also want the macros to be `IMPL`?

crtrott (approved):

crtrott (approved):

PR #152, opened 2025-10-02
Use Clang atomic min/max GCC-style builtins

dalg24:
Clang provides atomic min/max builtins with memory ordering as a language extension since LLVM 7.
https://releases.llvm.org/7.0.0/tools/clang/docs/LanguageExtensions.html#atomic-min-max-builtins-with-memory-ordering

This PR proposes to leverage them in the GCC-style fetch_op implementation.

Corresponds to https://github.com/kokkos/kokkos/pull/8507

PR #153, opened 2025-10-07
Fix pragma GCC diagnostic in Fetch_OP_GCC.hpp

masterleinad:
Fixes the `pragma push` clause in https://github.com/desul/desul/pull/152. We forgot to `push` the current values so that the `pop` would undo previous `push`es.


dalg24 (approved):
Thank you

PR #167, opened 2026-03-22
performance: Fix desul atomic regressions compared to RAJA for HIP backend

johnbowen42:
Using https://github.com/llnl/RAJA/blob/develop/benchmark/benchmark-atomic.cpp, I identified several cases where DESUL is slower than RAJA's atomic implementations for HIP, measured on MI250 and MI300 architectures.  This PR addresses those cases by using HIP's builtins for atomic store and load where available, and also uses HIP's atomic load builtin to short-circuit atomic max/min where no update is needed, especially helpful for atomics with low degrees of contention.

PR #171, opened 2026-04-06
Leverage HIP atomic load/store compiler built-ins

dalg24:
Extracting load/store performance fixes from #167 
The proposed approach is more in line with the intended way for specializing atomic operations in the backend.
I expect we will tackle fetch min and max in a follow up.
Note that I did not bother checking for existence of the compiler atomic builtin functions.
As far as I understand it, both Kokkos and RAJA only care about HIP versions that have them.


dalg24:
cc @johnbowen42

crtrott (approved):

dalg24:
@johnbowen42 @trws any objection to merging this?

johnbowen42:
> @johnbowen42 @trws any objection to merging this?

No objections, looks good to me

PR #172, opened 2026-04-17
Tweak reinterpret_cast to avoid warning

brian-kelley:
I'm proposing this small change to avoid a warning that g++ gives when T is itself a pointer to an 8-byte type:
``warning: casting 'long int*' to 'cas_t&' {aka 'long int&'} does not dereference pointer``

Here I'm just moving the dereference of ``dest`` to the outside of the cast, so it should never cause a change in behavior but it does fix this warning ([example link on godbolt](https://godbolt.org/z/Ydo5eT6c7)).

This has been tricky to replicate within desul itself but a user of Kokkos Kernels is seeing it come up in a real application.


dalg24 (approved):

crtrott (approved):

trws (approved):

dalg24:
Thank you Brian

PR #173, opened 2026-04-24
Fix more reinterpret_cast warnings

brian-kelley:
Anywhere we have to reinterpret_cast a user's ``T value`` to an integer ``cas_t``  of the same size, prefer
``*reinterpret_cast<cas_t*>(&value)`` over ``reinterpret_cast<cas_t&>(value)``.

This avoids more potential warnings of the form
```
casting 'T1*' to 'cas_t&' {aka 'T2&'} does not dereference pointer
```
that happens when T is a pointer type. All of these other examples were missed in #172.

``Compare_Exchange_MSVC.hpp`` and ``Compare_Exchange_SYCL.hpp`` files were already doing the casts this way, but I'm not sure if that was just a style difference or also to avoid warnings.

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
@@ -68,8 +68,8 @@ T device_atomic_compare_exchange(
     using cas_t =
         std::conditional_t<(sizeof(T) == 4), unsigned int, unsigned long long int>;
     cas_t return_val = atomicCAS(reinterpret_cast<cas_t*>(dest),
-                                 reinterpret_cast<cas_t&>(compare),
-                                 reinterpret_cast<cas_t&>(value));
+                                 *reinterpret_cast<cas_t*>(&compare),
+                                 *reinterpret_cast<cas_t*>(&value));
     return reinterpret_cast<T&>(return_val);
 #ifdef DESUL_CUDA_ARCH_IS_PRE_PASCAL
   } else if constexpr (std::is_same_v<T, float>) {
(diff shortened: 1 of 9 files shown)


atomics/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -68,8 +68,8 @@ T device_atomic_compare_exchange(
     using cas_t =
         std::conditional_t<(sizeof(T) == 4), unsigned int, unsigned long long int>;
     cas_t return_val = atomicCAS(reinterpret_cast<cas_t*>(dest),
-                                 reinterpret_cast<cas_t&>(compare),
-                                 reinterpret_cast<cas_t&>(value));
+                                 *reinterpret_cast<cas_t*>(compare),
+                                 *reinterpret_cast<cas_t*>(value));
```
masterleinad: Doesn't that need to be

```suggestion
                                 *reinterpret_cast<cas_t*>(&compare),
                                 *reinterpret_cast<cas_t*>(&value));
```
(similarly elsewhere)

brian-kelley: 🤦 Yes, fixed this now.

brian-kelley:
Since I'm not sure if the PR testing covers all these cases, I ran the desul tests with openmp+cuda locally and also the kokkos core tests for good measure.

masterleinad (approved):
It would be good to have the correpsonding Kokkos pull request up before merging for better testing.

dalg24 (approved):

jczhang07:
@brian-kelley could you also update `kokkos/master`?

brian-kelley:
@jczhang07 I can't make PRs directly into master because it always points at the most recent release. @dalg24 opened a PR to kokkos/develop (see above). This should make into 5.2.0 or 5.1.2 (whichever comes first), and then kokkos/master will include it.

PR #176, opened 2026-06-09
Use pre-commit to tighten the formatting

JBludau:
I think it is just easier to use than making sure someone has a `clang-format 13` locally. Additionally, it still allows users to use their old way of doing things.

This also includes `cmake-format` as I think we should have formatting on all files.
Since I could not find a reason to not only run it on `include/atomics`, I ran the formatting on everything.
We could do more, like checking for the license in the files, but I wanted to keep it minimal.

Probably easiest to review the commits separately


dalg24 (commented):
This is a good addition.  Please rebase/resolve the conflict.

JBludau:
> This is a good addition. Please rebase/resolve the conflict.

done

dalg24 (approved):

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,16 @@
+name: pre-commit
+
+on:
+  workflow_call:
```
dalg24: This is never going to trigger with that setting

dalg24 (changes requested):

.github/workflows/pre-commit.yml
JBludau: yeah, just saw that and fixed in https://github.com/desul/desul/pull/176/commits/807b342869516d2e99c9a6505de0f828cfc6c898

dalg24 (approved):

PR #177, opened 2026-06-09
Split `load` and `store` operations so they can be implemented with special intrinsics

JBludau:
This is part of a general redesign to fix the semantic violation described in #164 and the performance problem identified in https://github.com/desul/desul/pull/167.

The main idea is to separate load and store and not implement them with CAS. This can then be used in the impl for `min` and `max` as a next step.

BUT: with this switch, we basically can only provide lock-free atomics if we have `load`, `store` and `cas` for that size. This means, that 128bit atomics (x86, hopper, and arm only provide `cas` but no `load` or `store`) will go through lock tables.

Corresponding Kokkos PR: https://github.com/kokkos/kokkos/pull/9219

### DIFF
--- a/atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
+++ b/atomics/include/desul/atomics/Compare_Exchange_CUDA.hpp
@@ -10,6 +10,7 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #define DESUL_ATOMICS_COMPARE_EXCHANGE_CUDA_HPP_
 
 #include <desul/atomics/Common.hpp>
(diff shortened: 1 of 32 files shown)


JBludau:
~~I did not do the formatting, but I was hoping for #176 to land first~~

masterleinad (approved):

crtrott (dismissed):

JBludau:
@trws could you give this a look and see if you guys are fine with this direction?
OpenACC and MSVC can be done in a follow up

trws:
I'm good with the direction, in fact I rather like it.

crtrott (approved):

PR #181, opened 2026-08-31
Relax RDC requirement 

JBludau:
enabling rdc downstream in a non-rdc desul should work as we need to call the ensure_lock_arrays_on_device anyways.

### DIFF
--- a/atomics/include/desul/atomics/Lock_Array_CUDA.hpp
+++ b/atomics/include/desul/atomics/Lock_Array_CUDA.hpp
@@ -61,11 +61,15 @@ void finalize_lock_arrays_cuda();
 /// That is the purpose of the ensure_cuda_lock_arrays_on_device function.
 #ifdef DESUL_ATOMICS_ENABLE_CUDA_SEPARABLE_COMPILATION
 extern
+#else
+static
 #endif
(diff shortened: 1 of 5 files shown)


NamjaeChoi:
It looks like for SYCL, the corresponding variables are already marked as static. So we can lift SYCL restriction as well?

JBludau:
> It looks like for SYCL, the corresponding variables are already marked as static. So we can lift SYCL restriction as well?

I am still thinking about this one. AFAIK with SYCL rdc only works in shared libraries, but I am not sure how that goes together in the end

masterleinad:
> I am still thinking about this one. AFAIK with SYCL rdc only works in shared libraries, but I am not sure how that goes together in the end

We are testing with SYCL+RDC using static libraries in our CI while it should also work with shared libraries (as the comments suggest) but we are not testing that currently.

JBludau:
> > I am still thinking about this one. AFAIK with SYCL rdc only works in shared libraries, but I am not sure how that goes together in the end
> 
> We are testing with SYCL+RDC using static libraries in our CI while it should also work with shared libraries (as the comments suggest) but we are not testing that currently.

Thanks @masterleinad for correcting my misconception I was just starting to search the docs about that.
But ultimately it will depend on if Kokkos also wants to allow disabling the constant launch to be usable

PR #184, opened 2026-09-10
NextSilicon: add atomics support for NextSilicon targets

kc9jud:


CMakeLists.txt
```diff
@@ -13,8 +13,8 @@ if(NOT BLT_CXX_STD)
       CACHE STRING "")
 endif()
 
-set(DESUL_DEPENDENT_OPTIONS ENABLE_OPENMP ENABLE_CUDA ENABLE_HIP ENABLE_SYCL
-                            ENABLE_TESTS)
+set(DESUL_DEPENDENT_OPTIONS ENABLE_OPENMP ENABLE_CUDA ENABLE_HIP
+                            ENABLE_NEXTSILICON ENABLE_SYCL ENABLE_TESTS)
```
dalg24: Look at L30

kc9jud: Done.

atomics/include/desul/atomics/Compare_Exchange.hpp
```diff
@@ -11,9 +11,16 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Macros.hpp>
 
-#ifdef DESUL_HAVE_GCC_ATOMICS
+// FIXME_NEXTSILICON: NextSilicon host atomics replace GCC; remove when
+// divergence between GCC and NextSilicon host atomics definitions is eliminated.
+#ifdef DESUL_HAVE_NEXTSILICON_ATOMICS
+#include <desul/atomics/Compare_Exchange_NextSilicon_Host.hpp>
+#elif defined DESUL_HAVE_GCC_ATOMICS
```
dalg24: For consistency
```suggestion
#elif defined(DESUL_HAVE_GCC_ATOMICS)
```

atomics/include/desul/atomics/Fetch_Op.hpp
```diff
@@ -11,9 +11,16 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Macros.hpp>
 
-#ifdef DESUL_HAVE_GCC_ATOMICS
+// FIXME_NEXTSILICON: NextSilicon host atomics replace GCC; remove when
+// divergence between GCC and NextSilicon host atomics definitions is eliminated.
+#ifdef DESUL_HAVE_NEXTSILICON_ATOMICS
+#include <desul/atomics/Fetch_Op_NextSilicon_Host.hpp>
+#elif defined DESUL_HAVE_GCC_ATOMICS
```
dalg24:
```suggestion
#elif defined(DESUL_HAVE_GCC_ATOMICS)
```

atomics/include/desul/atomics/Load_And_Store.hpp
```diff
@@ -11,9 +11,16 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 #include <desul/atomics/Macros.hpp>
 
-#ifdef DESUL_HAVE_GCC_ATOMICS
+// FIXME_NEXTSILICON: NextSilicon host atomics replace GCC; remove when
+// divergence between GCC and NextSilicon host atomics definitions is eliminated.
+#ifdef DESUL_HAVE_NEXTSILICON_ATOMICS
+#include <desul/atomics/Load_And_Store_NextSilicon_Host.hpp>
+#elif defined DESUL_HAVE_GCC_ATOMICS
```
dalg24:
```suggestion
#elif defined(DESUL_HAVE_GCC_ATOMICS)
```

atomics/include/desul/atomics/Lock_Free_Types_NextSilicon.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace Impl {
+
+// NextSilicon uses GCC __atomic builtins on-device with the same lock-free
+// rules as the host.
+template <class T>
+inline constexpr bool device_atomic_always_lock_free<T, void> =
+#ifndef DESUL_HAVE_LIBATOMIC
+    ((sizeof(T) == 1 && alignof(T) == 1) || (sizeof(T) == 2 && alignof(T) == 2) ||
+     (sizeof(T) == 4 && alignof(T) == 4) || (sizeof(T) == 8 && alignof(T) == 8)
+#ifdef DESUL_HAVE_16BYTE_LOCK_FREE_ATOMICS_DEVICE
```
dalg24: Instead of `_HOST` with GCC, got it.

atomics/include/desul/atomics/Adapt_NextSilicon.hpp
```diff
@@ -0,0 +1,15 @@
...
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_ADAPT_NEXTSILICON_HPP_
+#define DESUL_ATOMICS_ADAPT_NEXTSILICON_HPP_
+
+// NextSilicon uses GCC __atomic builtins with the same memory-order constants.
+#include <desul/atomics/Adapt_GCC.hpp>
```
dalg24: Unless we define something like
```C++
template <class MemoryOrder>
struct NextSiliconMemoryOrder : private GCCMemoryOrder<MemoryOrder><MemoryOrder>{
  using GCCMemoryOrder<MemoryOrder>::value;
};
```
if we are directly going to refer to `Impl::GCCMemoryOrder` in other files, then I would just drop that header and have them just include `#include <desul/atomics/Adapt_GCC.hpp>`

atomics/include/desul/atomics/Lock_Free_Types_NextSilicon_Host.hpp
dalg24, line 1: Never included

atomics/include/desul/atomics/Fetch_Op_NextSilicon_Host.hpp
```diff
@@ -0,0 +1,55 @@
...
+COPYRIGHT file for details.
+Source: https://github.com/desul/desul
+
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_FETCH_OP_NEXTSILICON_HOST_HPP_
+#define DESUL_ATOMICS_FETCH_OP_NEXTSILICON_HOST_HPP_
+
+#include <desul/atomics/Adapt_NextSilicon.hpp>
```
dalg24:
```suggestion
#include <desul/atomics/Adapt_GCC.hpp>
```

dalg24 (dismissed):
With
https://github.com/desul/desul/blob/6056b2313fb1658ce37379b1fb95873b5a681751/atomics/include/desul/atomics/Macros.hpp#L163-L167
none of the device code is ever used.

From all the source files, keep `Fetch_Op.hpp` and `Fetch_Op_NextSilicon_Host.hpp`, drop everything else, and rename `Fetch_Op_NextSilicon_Host.hpp` to `Fetch_Op_NextSilicon.hpp`.

kc9jud:
@dalg24 vastly simplified

masterleinad (approved):

JBludau (approved):
Ok, so everything goes through the GCC (host) atomics even though nextsilicon identifies as a device in Kokkos. I can not speak to the `is_arithmetic` specialization, but I guess there is documentation stating that this is the constraint.

The change will not give nice backtraces as it will say `host` in desul's function name even when run on device, but I think for atomics this is kind of acceptable.

I think it is worth adding a comment about host atomics being used on device for nextsilicon above this block:
https://github.com/desul/desul/blob/6056b2313fb1658ce37379b1fb95873b5a681751/atomics/include/desul/atomics/Macros.hpp#L163-L167

kc9jud:
> I think it is worth adding a comment about host atomics being used on device for nextsilicon above this block:

Agreed, updated.

cwpearson (approved):

atomics/include/desul/atomics/Macros.hpp
```diff
@@ -120,6 +121,7 @@ static constexpr bool desul_impl_omp_on_host() { return false; }
     DESUL_IMPL_STRIP_PARENS(CODE)           \
   }
 #else
+// NextSilicon atomics use host atomics on both host and device
```
JBludau: This needs to go to line 164 (it is currently in the `OPENMP_ATOMICS` block here which means OpenMP-Target).

Also I would rephrase to
`This is the fallback that only calls host code. Note: Nextsilicon devices also use host atomics`

JBludau (changes requested):
I think the comment position is wrong

atomics/include/desul/atomics/Macros.hpp
kc9jud: done

JBludau (approved):

crtrott (approved):

JBludau:
@dalg24 @trws can you guys take a look?

crtrott (approved):

dalg24 (dismissed):
I prefer explicit configuration option and macro defined in the configuration file to avoid ODR issues.

kc9jud:
> I prefer explicit configuration option and macro defined in the configuration file to avoid ODR issues.

@dalg24 what should be the meaning of the config option and what should the macro gate? We support these atomics on both host and device, and we don't do separate compilation for host vs. device.

dalg24:
> > I prefer explicit configuration option and macro defined in the configuration file to avoid ODR issues.
> 
> @dalg24 what should be the meaning of the config option and what should the macro gate? We support these atomics on both host and device, and we don't do separate compilation for host vs. device.

The cmake machinery you had in the initial version.
And in the code you have now, guard with `DESUL_HAVE_NEXTSILICON_ATOMICS` instead of `__NEXTSILICON__`

kc9jud:
Ah okay... is the plan to split out all the different host compilers to have their own versions of the headers? GCC/Clang/icpx/etc.? Otherwise are there any other cases where host compilers get their own headers?

And do you want to duplicate the GCC headers too?

crtrott:
I don't think Damien means to introduce new headers. Just a configure option instead of auto-detecting NExtSilicon compilation. That said I don't see what kind of ODR violation one would get. I also don't see how this is different from use of pragmas such as `#if defined(__GNUC__) && (!defined(__clang__))`. We don't have a configure option for compiling with Clang vs compiling with GNU either.

crtrott:
Here is a blob which does the same kind of specialization difference for Intel compiler for example:

```
#if defined(__clang__)
#if (__clang_major__ >= 17) && \
    (!defined(__INTEL_LLVM_COMPILER) || __INTEL_LLVM_COMPILER >= 20240000)
// the suppression is not necessary from Clang 19 onwards
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Watomic-alignment"
DESUL_IMPL_GCC_HOST_ATOMIC_FETCH_OP(_min, arithmetic_not_long_double)
DESUL_IMPL_GCC_HOST_ATOMIC_FETCH_OP(_max, arithmetic_not_long_double)
#pragma GCC diagnostic pop
#else
DESUL_IMPL_GCC_HOST_ATOMIC_FETCH_OP(_min, std::is_integral)
DESUL_IMPL_GCC_HOST_ATOMIC_FETCH_OP(_max, std::is_integral)
#endif
#endif
```

So I need to hear more reasons for adding a cmake option, for now I don't want one.

PR #186, opened 2026-09-16
MSVC bug fix: 16-byte atomic exchange using lock table

dalg24:
Assisted-by: Codex:gpt-5.6-terra

Corresponds to https://github.com/kokkos/kokkos/pull/9553

compare exchange on 16-byte data types are not supposed to go through the lock arrays, or they may race with other atomic operations that are lock free.


atomics/include/desul/atomics/Compare_Exchange_MSVC.hpp
```diff
@@ -68,9 +68,24 @@ std::enable_if_t<sizeof(T) == 8, T> host_atomic_exchange(T* const dest,
 }
 
 template <class T, class MemoryOrder, class MemoryScope>
-std::enable_if_t<(sizeof(T) != 1 && sizeof(T) != 2 && sizeof(T) != 4 && sizeof(T) != 8),
-                 T>
-host_atomic_exchange(T* const dest, T val, MemoryOrder, MemoryScope scope) {
+std::enable_if_t<!host_atomic_always_lock_free<T>, T> host_atomic_exchange(
```
JBludau:
```suggestion
std::enable_if_t<sizeof(T) == 16, T> host_atomic_exchange(
```

JBludau (commented):
I think we need to explicitly enable it for 16byte.

Another solution to all of this would be to remove the 16byte atomics altogether, since we do not have a load and store for them...

JBludau (approved):
definitely a good bug fix.

crtrott:
What does load/store do for 16byte on windows?

crtrott (approved):
Just for the record load/store goes through compare_exchange which at least for the load from "const" values is via a UB impl. But we lived with that for a while.

